PR #3, opened 2020-06-18
OpenTelemetry Collector for Honeycomb

puckpuck:
chart to deploy the OpenTelemetry collector configured with recommended defaults and the Honeycomb exporter

### DIFF
--- a/opentelemetry-collector/Chart.yaml
+++ b/opentelemetry-collector/Chart.yaml
@@ -7,11 +7,11 @@ appVersion: 0.4.0
 keywords:
   - observability
   - tracing
-home: https://honeycomb.io
+home: https://opentelemetry.io/
 sources:
   - https://github.com/honeycombio/opentelemetry-collector
   - https://github.com/honeycombio/opentelemetry-collector-contrib
-icon: https://www.honeycomb.io/wp-content/themes/honeycomb/assets/img/logo.svg
+icon: https://opentelemetry.io/img/logos/opentelemetry-horizontal-color.png
 maintainers:
   - name: puckpuck
     email: pierre@honeycomb.io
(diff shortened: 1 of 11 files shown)


opentelemetry-collector/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v2
+name: opentelemetry-collector
+description: OpenTelemetry Collector for Honeycomb
+type: application
+version: 0.1.0
+appVersion: 0.4.0
+keywords:
+  - observability
+  - tracing
+home: https://honeycomb.io
```
irvingpop, line 10: is this the right homepage?

opentelemetry-collector/Chart.yaml
```diff
@@ -0,0 +1,17 @@
...
+sources:
+  - https://github.com/honeycombio/opentelemetry-collector
+  - https://github.com/honeycombio/opentelemetry-collector-contrib
+icon: https://www.honeycomb.io/wp-content/themes/honeycomb/assets/img/logo.svg
```
irvingpop, line 14: probably should be the OTel logo, right?

opentelemetry-collector/templates/NOTES.txt
```diff
@@ -0,0 +1,11 @@
+{{- $validApiKey := and (.Values.honeycomb.apiKey) (ne .Values.honeycomb.apiKey "YOUR_API_KEY") -}}
+
+{{- if not $validApiKey }}
+##### ERROR:
```
irvingpop, line 4: naive question, what happens if you hit this error condition? 

also if this is helpful/obvious to users, should we also require dataset be set?

irvingpop:
few minor things, other than that this looks great!

opentelemetry-collector/templates/NOTES.txt
puckpuck: dataset has a default name, and the Readme does explain how to change it, though we may want to make that part more prominent. If you don't specify an API key, the collector will throw unauthorized errors (via Honeycomb exporter).

PR #4, opened 2020-06-22
release otel-collector 0.1.0 chart

puckpuck:
release chart to helm repo


opentelemetry-collector-0.1.0.tgz
```diff
@@ -0,0 +1,88 @@
+apiVersion: v1
```
irvingpop, line 1: knowing very little about this process, but this looks like a YAML file with an extension .tgz

puckpuck: redid the release script and steps and it should be a YAML update + tgz binary now.

Ultimately releasing chart binaries needs to get automated.

puckpuck:
Will redo this release.  Found a small issue in readme.

PR #7, opened 2020-07-15
checksums for config and secrets

puckpuck:
Added checksum annotation(s) to all deploy pods. Updating a config setting or API key will now force a restart on all affected pods with a `helm upgrade ...` command.


puckpuck:
This will also address #9

PR #11, opened 2020-07-30
Secure Tenancy chart

puckpuck:
Helm chart to install Secure Tenancy.  Requires user to build and supply an image on private repository for Secure Tenancy binary.  

- Easy to get going, by default it will setup and install a MySQL database to use
- Can also use an existing MySQL database instead
- Migrate script will run with `helm install` commands, as either a post-install (when install MySQL) or pre-install (when using an existing MySQL) hook.
- Includes support for Ingress
- Includes support for horizontal pod autoscaling


nathanleclaire:
Beautiful! LGTM

PR #15, opened 2020-08-17
Add codeowners file

MikeGoldsmith:
Added CODEOWNERS file so team members are automatically set as reviewers. I've provisionally set @puckpuck as the only entry. Ideally it would be a team, but not sure if there is an appropriate one right now.

@paulosman @maplebed who do you think should be responsible for this?


MikeGoldsmith:
PS - I applied the _normal_ open source guidelines to this repo so it requires approval before merging and only allows 'squash and merge'.

PR #16, opened 2020-08-22
Add sample_rate and update version of otel collector

easel:
Adds sample_rate to the collector configuration and bumps the release to the latest.


easel:
Thanks for the merge!

PR #21, opened 2021-01-20
Always create ConfigMap

sarahhodne:
Currently the ConfigMap is only created when rbac.create=true, even though it's not an RBAC resource.

We've installed the chart with `rbac.create=false` to manage the RBAC resources out-of-band, but this caused the ConfigMap not to get installed. I can't think of a reason why the ConfigMap should be gated on `rbac.create`, or a reason to gate it on anything at all, but I'm happy to add a `configmap.create` or similar value if that would be useful.


puckpuck (approved):

puckpuck:
Thanks for this.

PR #24, opened 2021-03-03
Updates version for all helm charts

puckpuck:
Honeycomb agent -> 2.1.2
Open-Telemetry Collector -> 0.21.0
Secure Tenancy -> 1.6.1

Chart versions were also bumped up accordingly.

Otel-collector 0.21 no longer has the queued_retry processor, subsequently it was removed from the default pipeline in the helm chart config.

### DIFF
--- a/opentelemetry-collector/templates/configmap.yaml
+++ b/opentelemetry-collector/templates/configmap.yaml
@@ -22,7 +22,6 @@ data:
         api_url: {{ .Values.honeycomb.apiHost }}
         dataset: {{ .Values.honeycomb.dataset }}
         sample_rate: {{ .Values.honeycomb.sample_rate }}
-        sample_rate_attribute: {{ .Values.honeycomb.sample_rate_attribute }}
     {{- if .Values.config.exporters }}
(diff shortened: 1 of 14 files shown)


opentelemetry-collector/templates/configmap.yaml
```diff
@@ -22,6 +22,7 @@ data:
         api_url: {{ .Values.honeycomb.apiHost }}
         dataset: {{ .Values.honeycomb.dataset }}
         sample_rate: {{ .Values.honeycomb.sample_rate }}
+        sample_rate_attribute: {{ .Values.honeycomb.sample_rate_attribute }}
```
McSick, line 25: Does this matter if the sample_rate_attribute by default is commented out?

puckpuck: It doesn't matter, a blank will be output, and the exporter's default will handle it, but I did want to remove the sample_rate_attribute property from this PR, and clearly I missed an entry.

McSick (approved):

PR #29, opened 2021-03-11
Automate chart releases

puckpuck:
This PR makes release Helm charts as simple as clicking the Merge button.

Currently releasing a chart requires someone to run a script, checkout gh-pages, copy some files, and issue a second PR.  If trying to release multiple charts at once, this requires more manual hand holding.

With the changes, all of that manual work goes away in the name of automation being a merge to main. Makes use of CircleCI automation and leverages the Helm [chart-releaser](https://github.com/helm/chart-releaser) utility to automate releases. 

When a merge to main is done: 
1. Will check for changes to any charts in the `charts/` folder
2. Generate Helm chart packages for changed charts
3. Publish chart packages as a GitHub release
4. Update the Helm repo on the `gh-pages` branch to reflect the new chart package

If a chart is updated, and the chart's version is not updated, the automation will fail and issue a notification.

The CircleCI project requires 3 environment variables to be setup:
- CR_TOKEN - a GitHub access token with permissions to create releases, and push to a branch
- GIT_EMAIL - the email address that will be associated with the release and helm repo update
- CHARTS_REPO - the url for the Helm charts repository -> `https://honeycombio.github.io/helm-charts`

A change to the folder structure for the repo was required for this script. As such external references to the Helm chart GitHub repos will need to be updated accordingly. There is a single reference in the Honeycomb Kubernetes docs that will need to be updated to the new path.  /cc @mjingle


MikeGoldsmith (changes requested):
I think this is a great step forward but would like to see publishing the charts based in a git tag instead if merging to master. This would allow multiple changes to be rolled up into a single release as required and would decouple the need for updating versions as part of a PR to add a feature / fix a bug.

The libhoney & beeline projects do releases based on tags, an example:
https://github.com/honeycombio/libhoney-go/blob/main/.circleci/config.yml#L112

puckpuck:
I updated the release script and CircleCi config to only run on releases named: `vX.Y.Z-chartname` where chartname is the name of the Helm chart to get released.  The script will pickup the chartname from the tag, and package/release only that chart for the Helm repo.

With this, we will end up with 2 releases in GitHub releases.  The first to trigger the actual release script, the second will be what the script generates as part of its release process.  

/cc @MikeGoldsmith

vreynolds (approved):

PR #30, opened 2021-03-16
Create config options for ports in values file

matoszz:
Creates a ports section / mappings to allow for each section to be toggle on/off, as well as the ports themselves to change. Modifies both the service and deployment

### DIFF
--- a/opentelemetry-collector/values.yaml
+++ b/opentelemetry-collector/values.yaml
@@ -83,6 +83,33 @@ service:
   type: ClusterIP
   annotations: {}
 
+ports: # configures both service, and deployment
+  otlp:
+    enabled: true
+    containerPort: 55681 # new port could also be 4317
+    servicePort: 55681
+    protocol: TCP
+  otlp-http:
+    enabled: true
+    containerPort: 55681 # http server starts up on 55681
+    servicePort: 55681
+    protocol: TCP
+  jaeger-thrift:
+    enabled: true
+    containerPort: 14268
+    servicePort: 14268
+    protocol: TCP
+  jaeger-grpc:
+    enabled: true
+    containerPort: 14250
+    servicePort: 14250
+    protocol: TCP
+  zipkin:
+    enabled: true
+    containerPort: 9411
+    servicePort: 9411
+    protocol: TCP
+
 # Resource memory limits should be modified in conjunction with memory_limiter processor
 resources: 
   limits:
--- a/opentelemetry-collector/values.yaml
+++ b/opentelemetry-collector/values.yaml
@@ -86,18 +86,33 @@ service:
 ports: # configures both service, and deployment
   otlp:
     enabled: true
-    containerPort: 55681 # new port could also be 4317
-    servicePort: 55681
+    containerPort: 55680 # "legacy" port for backwards compatibility
+    servicePort: 55680
     protocol: TCP
   otlp-http:
     enabled: true
     containerPort: 55681 # http server starts up on 55681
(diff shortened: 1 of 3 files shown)


opentelemetry-collector/values.yaml
```diff
@@ -83,6 +83,33 @@ service:
   type: ClusterIP
   annotations: {}
 
+ports: # configures both service, and deployment
+  otlp:
+    enabled: true
+    containerPort: 55681 # new port could also be 4317
```
puckpuck, line 89: I believe this should be `4317` for GRPC.  We should also add the legacyGRPC port of `55680` since many are still using it, and the collector will create the legacy port on startup if OTLP GRPC is configured.
https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/otlp.go#L130

opentelemetry-collector/values.yaml
```diff
@@ -83,6 +83,33 @@ service:
...
+    enabled: true
+    containerPort: 14268
+    servicePort: 14268
+    protocol: TCP
+  jaeger-grpc:
+    enabled: true
+    containerPort: 14250
+    servicePort: 14250
+    protocol: TCP
+  zipkin:
```
puckpuck, line 107: This may also be an opportunity to add the other not so popular Jaeger ports/formats for `thrift_binary` and `thrift_compact` which are ports: 6832 and 6831 respectively.
https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/jaegerreceiver/README.md

opentelemetry-collector/values.yaml
matoszz: do you want to own trying to maintain all the port mappings? the struct in the values file can be extended arbitrarily, so users can simply create / append new blocks as needed given it's a `range` over the top level and maps it out by `$key`

opentelemetry-collector/values.yaml
matoszz: mentioned this above, but the more common pattern (from my experience) is to put stock values / examples in the `values` file of the chart but make it an array or method that could be extended. This PR for my purposes is such that I'm creating a values entry in a private repo and pinning this chart to this uptream, where I can add name / port mappings that can override and/or extend the defaults provided

puckpuck: What we want to make sure is all the default ports that the collector will listen on for OTLP, Jaeger, and Zipkin are enabled by default when someone does a simple install using `helm install ... --set ...` (no values file supplied)

The collector has long been an an odd state, where these 3 receivers are enabled by default but not all ports to use them are opened for it.

opentelemetry-collector/values.yaml
puckpuck: Right now, this PR will break existing setups that expect port 55680 to be opened by default for OTLP GRPC. So most certainly we need to add the legacyGRPC port.

Since OpenTelemetry is moving to 4317 for GRPC, we should adopt that here as well to be forward compatible.

puckpuck (approved):
I made a small change to the port names to accommodate 15 character limits.  All else works great.

matoszz:
thanks for the quick review @puckpuck

matoszz:
@puckpuck do you need to create a new release version for this config to be used? I'm weirdly not seeing it leveraging the same chart version

puckpuck:
@matoszz the current way this repo is setup, it does require another PR to create a chart release. We are hoping to make this automated very soon with #29

PR #32, opened 2021-03-30
release new packed otel collector

matoszz:
Hello - I had created a [PR](https://github.com/honeycombio/helm-charts/pull/30) for some port configuration options related to the otel collector, and we have our charts setup internally to pin to the upstream (instead of doing a `helm install`) which means to get the new configuration options, I need the bundled up charts in the index to pull successfully and apply the configuration options. 

I believed I followed all the steps based on historical PR's in this repo, but didn't want to assume there should be a minor version increment over simply packaging up the current changes in main. Happy to bump the minor release version also. Let me know if there's anything missing.


puckpuck:
Our Helm chart release process is working and the opentelemetry-collector chart was just released through that process.  You can now get the latest chart after doing a `helm repo update` 

Closing this PR.

PR #40, opened 2021-05-03
[refinery] - adjust limits for production

puckpuck:
Adjusts limits to be in line with what is expected in production deployments.

Replica Count -> 3
Memory Limit -> 2GB
InMemoryCollector.MaxAlloc -> (85% of 2GB)
TraceTimeout -> 60s

This also adds lots of docs for each setting inside of the primary values.yaml file allowing users an easier way to understand each setting, and how to build their own values file.

### DIFF
--- a/charts/refinery/README.md
+++ b/charts/refinery/README.md
@@ -96,6 +96,20 @@ rules:
     AdjustmentInterval: 15
     Weight: 0.5
 ```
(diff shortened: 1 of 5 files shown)


robbkidd (approved):
Makes sense to me. 👍 

Docs are cool.

doc martin boots

PR #63, opened 2021-10-03
[refinery] update ingress support for k8s >= 1.19

### LINKED ISSUE #54: [refinery] - update ingress for k8s 1.22
Update ingress spec to work with Kubernetes 1.22.

In K8s 1.19 official support for the ingress API was added.  In 1.22 support for the beta API was removed.

puckpuck:

Starting with Kubernetes 1.19, the ingress controller was made stable. In Kubernetes >= 1.20 a warning will be displayed if the beta is used, with 1.22 marking when the beta is deprecated. 

This PR defines 2 ingresses that will be installed exclusively of each other. Depending on the version of the target Kubernetes (>=1.19, or <1.19) the stable or beta version of the ingress will be installed.

To keep the ingress simple, and avoid potential configuration confusion, support for multiple paths has also been removed. Frankly, it never worked before, so this is a low risk removal.

fixes #54

### DIFF
--- a/charts/refinery/Chart.yaml
+++ b/charts/refinery/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 1.1.2
-appVersion: 1.4.1
+version: 1.2.0
+appVersion: v1.5.0
 keywords:
   - refinery
   - honeycomb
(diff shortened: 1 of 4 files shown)


vreynolds:
Getting further... had to add an ingress controller, with the right ingress class (nginx), and solve the LB in front of it not being able to expose a port on the host 😵

PR #70, opened 2021-10-08
Refinery/additional labels on resources

sbeginCoveo:

- applying additional labels on various resources

- More flexibility around labeling k8s resources. Useful with affinities and selectors


puckpuck (approved):
This is great, thank you very much for this.  We will look to do the same with our other Helm charts.

PR #85, opened 2021-12-03
[refinery] Force pods to recycle on rules changes

### LINKED ISSUE #83: Refinery pods do not restart on config change
We saw an issue recently with the Helm chart for Refinery where adding some new hashing fields to the config didn't trigger a restart, and resulted in weird behavior from the cluster that forced us to manually restart the Refinery pods.

**Versions**

Chart: 1.3.0
Helm: ArgoCD v2.0.3+8
Kubernetes cluster: EKS 1.18.20
Refinery Version:  "v1.6.1@sha256:e742af48b2ac1eb6b0116246780eb7e4942f0251152742990def677d853a129a"

**Steps to reproduce**

1. Add a new field to the `values.yml` override for hashing field config
2. Deploy the new config
3. Notice that Refinery pods do not restart to consume the new config

MikeGoldsmith:
### Summary
Pods are not recycled when making changes to refinery sampling rules to a deployed cluster using `helm upgrade`. Refinery live-reloads changes to rules.yaml and does not strictly need a restart for the changes to be applied.

Changes to config.yaml doe result in pods being recycled.

However, this make it look like changes have not been applied to the cluster when looking at cluster health. This change adds a new pod annotation for the rules.yaml which ensures pods are recycled when rules.yaml are changed.

- Closes #83 

### Changes
- Add new pod annotation that uses the checksum of rules.yaml

### DIFF
--- a/charts/refinery/templates/deployment.yaml
+++ b/charts/refinery/templates/deployment.yaml
@@ -15,6 +15,7 @@ spec:
     metadata:
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmap-config.yaml") . | sha256sum }}
+        checksum/rules: {{ include (print $.Template.BasePath "/configmap-rules.yaml") . | sha256sum }}
         {{- with .Values.podAnnotations }}
         {{- toYaml . | nindent 8 }}
         {{- end }}


puckpuck:
The referenced issue is to restart on config changes, but this is implementing something to modify on rules changes.  Rules are all hot reloadable, and thus we shouldn't be doing this change.

MikeGoldsmith:
Closing this for now as rules changed intentionally do not require pod recycling. We're going to make a docs update to describe how config and rules changes are handled instead.

PR #92, opened 2022-01-12
agent: prepare release 1.2.0

vreynolds:
chart for agent release [2.3.0](https://github.com/honeycombio/honeycomb-kubernetes-agent/releases/tag/v2.3.0)


JamieDanielson (approved):
I've been wondering if we should start including the new k8s agent version in the changelog, but could go either way since it's already in the chart that may be sufficient.

vreynolds:
Yeah, I'm not sure what's best. Maybe a link to the other release and no other notes? Like a regular dependency update would look?

PR #95, opened 2022-01-24
[Refinery] Introduce optional init container for Refinery deployment to ensure Redis connectivity

jijotj:
### Which problem is this PR solving?
Refinery pods crash looping until Redis instances are healthy. 

### Short description of the changes
This PR adds an optional init container for Refinery deployment to sleep until Redis instances are up and running. 
No breaking changes.


MikeGoldsmith:
Hey @jijotj - thank you for the PR.

We think that refinery itself should support better start-up races with redis and not require k8s to manage that interaction. I've created https://github.com/honeycombio/refinery/pull/382 to add retry behaviour to refinery.

Once that PR has been accepted, we'll roll a new release of refinery and helm-chart.

PR #106, opened 2022-01-28
[opentelemetry-collector] - update to collector 0.43.0

puckpuck:
Updates to the latest version of the OpenTelemetry collector (0.43.0)

Because of a change in executable, and we should be following best practices here anyhow, this also removes the `command` property, and instead just uses `args`, which will allow the base image to define the entrypoint.


MikeGoldsmith (approved):
Thanks @puckpuck 👍🏻

PR #110, opened 2022-02-03
Prepare refinery v1.5.0 release

MikeGoldsmith:
### Description
Prepares the Refinery v1.5.0 release.

### Changes
- Adds changelog entry
- Updates chart version to 1.5.0


JamieDanielson (commented):
If we’re doing a minor bump already, I’m thinking we should first update `appVersion` of Refinery in `Chart.yaml` to 1.9.0 in a separate PR, with the changes needed from the Refinery config updates (e.g. `REFINERY_GRPC_LISTEN_ADDRESS`). Once that’s merged in, this can be rebased to include that change and then should be good to go.

JamieDanielson (approved):
Looks good, might consider noting what changed in refinery with the updated version but not a blocker

PR #111, opened 2022-02-03
[refinery] Bump refinery image version to 1.9.0

MikeGoldsmith:
### Description
Bumps the refinery image version to latest.

### Changes
- Updates the refinery image version to 1.9.0


JamieDanielson (commented):
As we chatted about offline, we'll want to add in the config changes as well 👍

JamieDanielson (approved):

PR #112, opened 2022-02-09
docs: how to ingress local refinery chart

vreynolds:


MikeGoldsmith (approved):
Looks good.

Makes me wonder if we could / should look at creating a fork of an ingress controller (eg nginx) that could do trace ID assignment and route to correct host. It would take pressure off the nodes having to re-route traffic to peers.

bixu:
> Looks good.
> 
> Makes me wonder if we could / should look at creating a fork of an ingress controller (eg nginx) that could do trace ID assignment and route to correct host. It would take pressure off the nodes having to re-route traffic to peers.

Trace-aware ingress could also be useful for a pipeline like the one we are building at SumUp, where we have Otel Collectors in between ingress and Refinery. If we want to do any kind of head sampling in the Collector, we'd need a trace-aware load balancer in front.

CC @svrakitin

vreynolds:
@MikeGoldsmith and/or @bixu I'll leave it up to one of you to create an issue for the enhancement idea ❤️

PR #120, opened 2022-02-14
[Refinery] Remove resource requests and limits from helm default values

jijotj:
### Which problem is this PR solving?
Unset default CPU limit when installing Refinery as a sub chart. The usual convention of helm charts is to not have default resource requests/limits set. This allows releases to not set resource limits if needed (for ex: to avoid CPU throttling with CPU limits). While helm defaults can be unset by setting it to `null` as an override, this doesn't work when installing the chart as a sub chart. Unfortunately, this is an open bug in helm: https://github.com/helm/helm/issues/9136

### Short description of the changes
Remove resource requests and limits from helm default values

### DIFF
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -15,14 +15,14 @@ replicaCount: 3
 
 # Changing memory limits from the default 2Gi, requires updates to the
 # config.InMemCollector.MaxAlloc property.
-resources:
-  limits:
-    cpu: 2000m
+resources: {}
+  # limits: 
+    # cpu: 2000m
     # Changing memory limits requires updating config.InMemCollector.MaxAlloc
-    memory: 2Gi
-  requests:
-    cpu: 500m
-    memory: 500Mi
+  #   memory: 2Gi
+  # requests:
+  #   cpu: 500m
+  #   memory: 500Mi
 
 image:
   repository: honeycombio/refinery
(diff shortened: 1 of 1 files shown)


JamieDanielson:
Thanks for the detail @jijotj! This sounds like a frustrating issue! Ideally we’d like the fix from helm, but it looks like it’s been there for a while so I see why we need to find an alternative solution in the meantime.

We have these default values to help ensure limits are set for Refinery; without them, Refinery would end up eating all available resources and ultimately crashing, which will be unexpected especially for new users. Because of this, removing from the file is not really a viable solution.

Reviewing some of the comments on that helm issue, and some other linked issues for others experiencing the same inability to override values for sub-charts, it appears there may be a [couple of workarounds](https://github.com/helm/helm/issues/9136#issuecomment-1015864981) and I’d be curious to know if any of them work for you. If so, we may at least be able to add some documentation to help with this while it remains an open issue in helm.

When deploying, you could add a command-line argument like `--set resources.limits.cpu=false` or `--set resources.limits.cpu=newValue` where `newValue` represents the actual limit that should be set. Alternatively, you may also be able override in the [helm install](https://helm.sh/docs/helm/helm_install/) command with `-f` as in `helm install refinery honeycomb/refinery -f values.yaml -f override.yaml`. The last (right-most) value should take precedence.

jijotj:
Thanks for your response. I was able to manage with one of the workarounds. Closing this PR now.

PR #127, opened 2022-03-02
[honeycomb] - use 2.4.0 agent with retry buffer

puckpuck:
- Adds `retryBufferSize` and `retryBufferExpire` as properties to control their equivalent for the agent
- Adds `additionalFields` to the config map. It was missing before, which was a bug
- Updates app version to `2.4.0` to match agent version
- Bumps minor version of Helm chart

### DIFF
--- a/charts/honeycomb/Chart.yaml
+++ b/charts/honeycomb/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: honeycomb
 description: Honeycomb Kubernetes Agent
-version: 1.3.1
-appVersion: 2.3.2
+version: 1.4.0
+appVersion: 2.4.0
 keywords:
   - observability
   - tracing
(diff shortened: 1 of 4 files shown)


charts/honeycomb/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: honeycomb
 description: Honeycomb Kubernetes Agent
-version: 1.3.1
-appVersion: 2.3.2
+version: 1.4.0
```
MikeGoldsmith, line 4: We prefer to update the chart version as part of the release PR/commit that includes the changelog and is what is tagged.
```suggestion
version: 1.3.1
```

MikeGoldsmith (approved):
Looks good other than not updating chart version. We'll do that in release PR 👍🏻

PR #133, opened 2022-04-11
Update to latest agent

kentquirk:
honeycomb-kubernetes-agent has been updated.


JamieDanielson (approved):
Thanks Kent!
I've changed my mind on our earlier discussion about an example of the `exclude` option here. You've already [added to the docs](https://docs.honeycomb.io/getting-data-in/integrations/kubernetes/configuration/#log-watchers-configuration), and I'll add a note in the release. I notice the similar `paths` option isn't shown as an example here either so there isn't prior art on having this specific config in the quickstart. We'll revisit if it becomes a concern 👍

PR #135, opened 2022-04-14
Add ingressClassName to refinery

kentquirk:
This adds support to refinery for [ingressClassName](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation) which is a more recent feature of Kubernetes.


MikeGoldsmith (approved):

kentquirk:
This was redundant.

PR #137, opened 2022-04-18
[refinery] prep refinery chart v1.9.0

JamieDanielson:
**changes made**

- update changelog for refinery chart
- bump refinery chart version to v1.9.0


kentquirk (approved):
👍

PR #139, opened 2022-05-17
Bump refinery to 1.14.1

kentquirk:
Bump refinery version, as well as extend the demo yaml file

### DIFF
--- a/charts/refinery/sample-configs/rules_complete.yaml
+++ b/charts/refinery/sample-configs/rules_complete.yaml
@@ -220,17 +220,17 @@ dataset3:
 dataset4:
   Sampler: RulesBasedSampler
   # Optional, if set to true then the rules will also check nested json fields, in the format of parent.child
-	CheckNestedFields: false
-	dataset4.rule:
-		- name: "drop healthchecks"
+  CheckNestedFields: false
+  dataset4.rule:
+    - name: "drop healthchecks"
       drop: true
       condition:
         field: http.route
         operator: '='
         value: "/health-check"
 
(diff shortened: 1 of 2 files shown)


charts/refinery/sample-configs/rules_complete.yaml
```diff
@@ -21,6 +21,15 @@ Sampler: DeterministicSampler
 # Eligible for live reload.
 SampleRate: 1
 
+# With the change to support environemt and services in Honeycomb, some users will
+# want to support both sending telemetry to a legacy dataset and a new environment
+# called the same thing (eg `production`).
+#
+# This can be accomplished by leveraging the new `DatasetPrefix` configuration property
+# and then using that prefix in the rules definitions for the legacy datasets.
+# DatasetPrefix: "production"
```
MikeGoldsmith, line 30:
```suggestion
# With the change to support environemt and services in Honeycomb, some users will
# want to support both sending telemetry to a classic dataset and a new environment
# called the same thing (eg `production`).
#
# This can be accomplished by leveraging the new `DatasetPrefix` configuration property
# and then using that prefix in the rules definitions for the classic datasets.
# DatasetPrefix: "production"
```

MikeGoldsmith (commented):
Changes look good - just made one small correction in using classic over legacy. I'll check refinery docs incase we have the same error there.

I ran the file through a yaml validator and it failed. Please could you check it's good before merging 😄

MikeGoldsmith (approved):

robbkidd (approved):

PR #143, opened 2022-06-01
[oopentelemetry-collector] bump collector to 0.52.0

MikeGoldsmith:
Updates the OpenTelemetry chart app version to latest (0.52.0).


JamieDanielson (commented):
I'm thinking this should be at least a minor instead of a patch, because it's such a large jump from the previous version and there are several breaking changes. Not sure whether it warrants a major since the collector itself is still in 0. though

PR #147, opened 2022-06-07
[Refinery] Enable extra volume configurations in deployment config

mars64:

These changes enable users to define additional `volume` and `volumeMount` configurations for the Refinery deployment object. This is important to enable volume-based secrets handling techniques such as the [secrets-store-csi-driver](https://secrets-store-csi-driver.sigs.k8s.io/).

Adds two new helm values: `volumes: [ ]` and `volumeMounts: [ ]`, uses the existing patterns to call the values using `toYaml` and `nindent`.

From `./charts/refinery`, I am able to confirm that the chart renders properly _without_ the values set:

```
% helm template -f values.yaml .
```

and _with_ the values set:

```
% cat /tmp/volumes.values.yaml  
extraVolumeMounts:
  - name: honeycomb-secrets
    mountPath: "/mnt/secrets-store"
    readOnly: true

extraVolumes:
  - name: honeycomb-secrets
    csi:
      driver: secrets-store.csi.k8s.io
      readOnly: true
      volumeAttributes:
        secretProviderClass: "honeycomb-secrets"

% helm template -f values.yaml -f /tmp/volumes.values.yaml .
```

### DIFF
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -48,7 +48,7 @@ environment: [ ]
   #       name: honeycomb
   #       key: api-key
 
-# Use this in conjunction with `.Values.extraVolumes` to map additional volumes into the Refinery pods 
+# Use to map additional volumes into the Refinery pods 
 # Useful for volume-based secrets
 extraVolumeMounts: [ ]
   # - name: honeycomb-secrets
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -55,7 +55,7 @@ extraVolumeMounts: [ ]
   #   mountPath: "/mnt/secrets-store"
   #   readOnly: true
 
-# Use this in conjunction with `.Values.extraVolumeMounts` to map additional volumes into the Refinery pods 
+# Use to map additional volumes into the Refinery pods 
 extraVolumes: [ ]
   # - name: honeycomb-secrets
   #   csi:
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -48,6 +48,22 @@ environment: [ ]
   #       name: honeycomb
   #       key: api-key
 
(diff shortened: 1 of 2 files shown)


MikeGoldsmith:
Hi @mars64 - thanks for the PR and sorry it's taken so long to get back to you. I'll be reviewing shortly to provide feedback.

charts/refinery/values.yaml
```diff
@@ -48,6 +48,22 @@ environment: [ ]
   #       name: honeycomb
   #       key: api-key
 
+# Use this in conjunction with `.Values.extraVolumes` to map additional volumes into the Refinery pods
```
MikeGoldsmith, line 51:
```suggestion
# Use to map additional volumes into the Refinery pods 
```

charts/refinery/values.yaml
```diff
@@ -48,6 +48,22 @@ environment: [ ]
...
   #       key: api-key
 
+# Use this in conjunction with `.Values.extraVolumes` to map additional volumes into the Refinery pods 
+# Useful for volume-based secrets
+extraVolumeMounts: [ ]
+  # - name: honeycomb-secrets
+  #   mountPath: "/mnt/secrets-store"
+  #   readOnly: true
+
+# Use this in conjunction with `.Values.extraVolumeMounts` to map additional volumes into the Refinery pods
```
MikeGoldsmith, line 58:
```suggestion
# Use to map additional volumes into the Refinery pods 
```

MikeGoldsmith (changes requested):
Looks great, thanks @mars64 👍🏻 

I've left a couple of minor suggestions to tidy up the example text. We don't need the example wording to include how the values will be used in deployment.yaml template 😄

mars64:
Applied suggestions. Let me know if you need anything else from me here.

MikeGoldsmith (approved):
Great, thanks @mars64

PR #155, opened 2022-07-15
Add helm namespace to templates

### LINKED ISSUE #146: Update chart to use helm release namespace

We utilize the helm templating engine to generate our manifests which we then check in and apply via kubectl at a later time. Right now, this chart doesn't inject the helm release namespace into the resources' metadata. It instead relies on the underlying kubectl client having a namespace flag specified. 

A common pattern is to include this in the templates
```
metadata:
    namespace: {{ .Release.Namespace }}
```

This lets you still use the existing way of applying via helm or kubectl and passing the namespace flag, but also allows generating templates with the namespace dictated in the yaml. 

**Describe alternatives you've considered**

We could work around this issue by writing code so our deployment pipeline can take the namespace and apply it with kubectl. Other charts we use specify this namespace, so it would be nice if they were consistent.

**Additional context**
There was a long conversation about this [here](https://github.com/helm/helm/issues/5465). It was ultimately closed as they won't dictate one way or the other. There are lots of issues that link to it though where others have added in this config.

Let me know your thoughts on this particular matter. If you approve, we can do the PR. Thanks for your consideration!

alex-bezek:
- Closes #146

While helm install can respect the kubectl context namespace or accept a namespace flag, when using helm template, this information is lost if not persisted to the generated files. This means tooling that requires a helm template -> kubectl install has to work around the missing namespace.

Adds in the default `.Release.Namespace` value to the namespace for each namespace resource. https://helm.sh/docs/chart_template_guide/builtin_objects/ . Similar changes have been made such as https://github.com/helm/charts/pull/17937/files#diff-4dcbfca5f8fd1c82a179cf60e0448926548323d101f0321679e2fd1e3a2dc665R12  found on the original issue https://github.com/helm/helm/issues/5465

```
❯ kns
Context "ngrok-dev-us" modified.
Active namespace is "default".
❯ helm template ./charts/opentelemetry-collector | grep -i namespace
  namespace: default
  namespace: default
  namespace: default
  namespace: default
  namespace: default
❯ kns
Context "ngrok-dev-us" modified.
Active namespace is "kube-system".
❯ helm template ./charts/opentelemetry-collector | grep -i namespace
  namespace: kube-system
  namespace: kube-system
  namespace: kube-system
  namespace: kube-system
  namespace: kube-system

```

### DIFF
--- a/charts/opentelemetry-collector/templates/configmap.yaml
+++ b/charts/opentelemetry-collector/templates/configmap.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ include "opentelemetry-collector.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
   {{- include "opentelemetry-collector.labels" . | nindent 4 }}
 data:
--- a/charts/opentelemetry-collector/templates/deployment.yaml
+++ b/charts/opentelemetry-collector/templates/deployment.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "opentelemetry-collector.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "opentelemetry-collector.labels" . | nindent 4 }}
 spec:
--- a/charts/opentelemetry-collector/templates/hpa.yaml
+++ b/charts/opentelemetry-collector/templates/hpa.yaml
@@ -3,6 +3,7 @@ apiVersion: autoscaling/v2beta1
 kind: HorizontalPodAutoscaler
 metadata:
   name: {{ include "opentelemetry-collector.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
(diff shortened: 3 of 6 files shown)


kentquirk:
@alex-bezek Apologies for the delay. This looks good -- the main question before I approve it is whether there are any updates needed to the README file? I don't know enough to know if something there should change as a result of this.

alex-bezek:
No problem @kentquirk. This change is just leveraging an existing helm built in value the `Release` object. https://helm.sh/docs/chart_template_guide/builtin_objects/ . This particular one, the namespace, is set by helm. If you do a `helm install` it uses the namespace of your current kubecontext. If using `helm template` there is a `--namespace` flag. 

I wouldn't document the built-in objects in your readme, but if you'd like I could add just a section explaining that namespaces are added to all namespace-able objects and it uses the helm internals to do so.

kentquirk (approved):
Thank you for this!

kentquirk:
Sounds like we don't need any documentation changes. I just wanted to be sure that there wasn't something non-obvious.

PR #162, opened 2022-07-27
Custom rules configmap

teaguecole:

When creating sampling rules for a larger amount of datasets with Honeycomb Classic using the current method, the Values file gets larger and larger and harder to maintain. This PR is to enable users to pass in the name of their own configmap.

My particular use case is using Refinery as a subchart inside of a larger parent chart. If I have my own ConfigMap template, I will the be able to use templating to merge a collection of YAML within the `files/` directory of my parent chart. This will enable the possibility of a YAML file per dataset. All of the files will get merged in to one value for the key rules.yaml as expected.  However, it's up. to the user to figure this out.

The default behavior is to work as is.

- Closes #<enter issue here>

Added logic to only create a configmap for rules if the user doesnt add a name for a different configmap to use.  The name field for the rules configmap in the deployment is either the default or the name passed in the vaules file.

If the name for the custom configmap for rules is left blank you get

```
RulesConfigMapName:
```

`helm template .`

```
volumes:
- name: refinery-config
projected:
sources:
- configMap:
name: release-name-refinery-config
items:
- key: config.yaml
path: config.yaml
- configMap:
- name: release-name-refinery-rules
items:
- key: rules.yaml
path: rules.yaml
```

If it's set you get

`helm template .`

```
RulesConfigMapName: "customName"
```

```
volumes:
- name: refinery-config
projected:
sources:
- configMap:
name: release-name-refinery-config
items:
- key: config.yaml
path: config.yaml
- configMap:
- name: customName
items:
- key: rules.yaml
path: rules.yaml
```

### DIFF
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -230,9 +230,10 @@ config:
     # listen for requests for /metrics.
     MetricsListenAddr: 0.0.0.0:9090
 
-  # When Name is left blank, refinery sets up a ConfigMap with the rules defined below.
-  # The alternative is to pass in your own ConfigMap with the rules defined there.
-  ConfigMapRulesName:
+  # When RulesConfigMapName is left blank, refinery sets up a ConfigMap with the rules defined below.
+  # The alternative is to pass in your own ConfigMap with the rules defined there. The only caveat is 
+  # you need to use rules.yaml as the key, but the value can be templated out.
+  RulesConfigMapName:
 
 # Values used to build rules.yaml
 # See the example in sample-configs/rules_complete.yaml for full details on all properties
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -48,7 +48,7 @@ environment: [ ]
   #       name: honeycomb
   #       key: api-key
 
-# Use to map additional volumes into the Refinery pods
+# Use to map additional volumes into the Refinery pods 
 # Useful for volume-based secrets
 extraVolumeMounts: [ ]
   # - name: honeycomb-secrets
@@ -235,7 +235,7 @@ config:
   UseRulesConfigMap: true
 
   # If UseRulesConfigMap is false, set the name of your own ConfigMap to use.  
-  # CustomRulesConfigMapName: my-refinery-rules
+  # CustomRulesConfigMapName: 
 
 # Values used to build rules.yaml
 # See the example in sample-configs/rules_complete.yaml for full details on all properties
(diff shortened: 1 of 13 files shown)


charts/refinery/values.yaml
```diff
@@ -230,6 +230,11 @@ config:
     # listen for requests for /metrics.
     MetricsListenAddr: 0.0.0.0:9090
 
+  # When Name is left blank, refinery sets up a ConfigMap with the rules defined below.
+  # The alternative is to pass in your own ConfigMap with the rules defined there.
+  ConfigMap:
```
pkanal, line 235: Is there a reason that this needs to be nested 3 levels deep? Could we use something like `ConfigMapName` instead?

teaguecole: Not necessarily. I can make that change. :)

charts/refinery/values.yaml
```diff
@@ -230,6 +230,10 @@ config:
     # listen for requests for /metrics.
     MetricsListenAddr: 0.0.0.0:9090
 
+  # When RulesConfigMapName is left blank, refinery sets up a ConfigMap with the rules defined below.
+  # The alternative is to pass in your own ConfigMap with the rules defined there. The only caveat is 
+  # you need to use rules.yaml as the key, but the value can be templated out.
+  RulesConfigMapName:
```
MikeGoldsmith, line 236: Let's use `""` as a default value instead of `null`. It doesn't affect the if condition and looks nicer in the template output.

```suggestion
  RulesConfigMapName: ""
```

charts/refinery/values.yaml
```diff
@@ -230,6 +230,10 @@ config:
     # listen for requests for /metrics.
     MetricsListenAddr: 0.0.0.0:9090
 
+  # When RulesConfigMapName is left blank, refinery sets up a ConfigMap with the rules defined below.
+  # The alternative is to pass in your own ConfigMap with the rules defined there. The only caveat is 
+  # you need to use rules.yaml as the key, but the value can be templated out.
```
MikeGoldsmith, line 235:
```suggestion
  # RulesConfigMapName is used to override the default ConfigMap that defines the rules yaml.
  # When blank, refinery is configured using the a ConfigMap based on the rules below.
  # When set, refinery is configured using the rules defined in the provided ConfigMap. The ConfigMap must use rules.yaml as the key but the value can be templated.
```

MikeGoldsmith (changes requested):
Looks good - thanks @teaguecole. We really appreciate the contributions 👍🏻 

I've added a couple of minor tweaks the default value and description of the RulesConfigMapName field then we're good to merge.

MikeGoldsmith (approved):

MikeGoldsmith:
Thanks @teaguecole 🎉

teaguecole:
@MikeGoldsmith looking at the contribution docs and the history of releases, it appears that releases for refinery helm chart come out roughly every month?  Just trying to ballpark estimate when I can expect to see this come through.

MikeGoldsmith:
@teaguecole we don't have a set cadence for these tools, it's when we have something to release. I'll look to process a release soon.

PR #164, opened 2022-08-10
[opentelemetry-collector] Swap order of hpa spec.metrics

chrisdotm:
Really this is for https://github.com/argoproj/argo-cd/issues/1079 which points blame at upstream k8s.

https://github.com/argoproj/argo-cd/issues/1079
This issue points at upstream k8s as the problem, but for the moment it's easier to request a change here that matches the output of the HPA object

Swapping the CPU and memory metric for the HPA

Nothing should change other than the resulting order of the rendered template

### DIFF
--- a/charts/opentelemetry-collector/Chart.yaml
+++ b/charts/opentelemetry-collector/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: opentelemetry-collector
 description: OpenTelemetry Collector for Honeycomb
 type: application
-version: 1.2.1
+version: 1.2.0
 appVersion: 0.52.0
 keywords:
   - observability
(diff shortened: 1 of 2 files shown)


chrisdotm:
I wasn't sure if the tag was for me, so I committed the patch change, I can revert that if it wasn't for me

JamieDanielson:
> I wasn't sure if the tag was for me, so I committed the patch change, I can revert that if it wasn't for me

Oops, sorry about that confusion! That was for me 😅  Can you revert that change and I'll have the version updated in the release?

This looks good, just spent too long looking at yaml yesterday and wanted to make sure I was double checking with fresh eyes today. Thanks @chrisdotm !

JamieDanielson (approved):
Thanks @chrisdotm !

Appreciate you sharing the link to the details. Looks like this was changed in the [upstream collector chart](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/103) already in December 2021, and similar changes made in other helm charts like redis, nginx, etc. We'll get this in 👍

PR #165, opened 2022-08-16
Use correct syntax in commented out field

jazzdan:

Without this if you removed the comment in front of `RedisUsername` it would be a YAML syntax error.
<please describe the issue>


MikeGoldsmith (approved):
Yep, thanks for catching @jazzdan 👍🏻

PR #167, opened 2022-08-18
docs: document metrics.interval configuration option

bodymindarts:

We would like to reduce the amount of data ingested for our non-prod environments by reducing the number of metrics events. I had to go down a rabbit hole until I found [the code in the honeycomb agent](https://github.com/honeycombio/honeycomb-kubernetes-agent/blob/9601c83cb12ef9e8b443641b3111a749c4653645/config/config.go#L51) that showed it accepted a `metrics.interval` configuration.

This PR saves the next searcher that same trip.


MikeGoldsmith (approved):
Thanks for the contribution @bodymindarts - much appreciated 👍🏻

Issue #157, opened 2022-07-21
Add namespace to templates for Refinery


kentquirk:
#146 described a problem, and #155 fixed it, but only for the collector helm chart. We should have the same feature in the Refinery helm chart.

alex-bezek:
@kentquirk I can do this and https://github.com/honeycombio/helm-charts/issues/158 . Is it ok if I do them in a single PR?

Issue #158, opened 2022-07-21
Add namespace support to Honeycomb helm chart


kentquirk:
#146 described a problem, and #155 fixed it, but only for the collector helm chart. We should have the same feature in the Honeycomb helm chart.

pdeveltere:
Yes, would be lovely if we could specify the namespace for the honeycomb agent 🙏🏽

PR #175, opened 2022-09-01
Add Helm namespace metadata

masonjlegere:
This allows generating the namespace and have it defined based on the kubectl client's namespace flag

See Issue https://github.com/honeycombio/helm-charts/issues/146 for context around the problem

- Closes [#157](https://github.com/honeycombio/helm-charts/issues/157)

Please see the same PR for the Collector which makes the same changes: https://github.com/honeycombio/helm-charts/pull/155

### DIFF
--- a/charts/refinery/templates/configmap-config.yaml
+++ b/charts/refinery/templates/configmap-config.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ include "refinery.fullname" . }}-config
+  namespace: {{ .Release.Namespace }}
   labels:
   {{- include "refinery.labels" . | nindent 4 }}
 data:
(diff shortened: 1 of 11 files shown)


pkanal (commented):
Thank you for updating this, it looks good to me! ~It looks like we also want support for namespace in the Honeycomb helm chart (#158). Could you add that as part of this PR as well?~ On second thought, we can ship the Honeycomb chart changes separately.

Could you also add namespace support for the remaining templates in the refinery templates:
- [service-redis.yaml](https://github.com/honeycombio/helm-charts/blob/main/charts/refinery/templates/service-redis.yaml)
- [service.yaml](https://github.com/honeycombio/helm-charts/blob/main/charts/refinery/templates/service.yaml)
- [serviceaccount.yaml](https://github.com/honeycombio/helm-charts/blob/main/charts/refinery/templates/serviceaccount.yaml)

MikeGoldsmith (dismissed):
As per @pkanal request, please can you update the other templates to add the namespace value 👍🏻

masonjlegere:
Ops sorry forgot the other files - should be good now

kentquirk (approved):
Thank you.

PR #176, opened 2022-09-03
Calculate `MaxAlloc` based on pod memory request.

bixu:
If `config.InMemCollector.MaxAlloc` is set explictly, use that value instead.

It's tedious to manually-calculate the value of `MaxAlloc` when experimenting with Refinery memory tunings.

This PR doesn't address any specific GitHub issues, and is based on brief chats with various Honeycomb employees on public Slack. Keeping as a draft until feedback is recieved.

The chart version has not been bumped out of respect for Honeycomb's internal versioning process.

This PR aims to do the following:

- Calculate a safe value for `MaxAlloc` automatically
- Accept an explicit override if one is present.

Run

```
helm template charts/refinery | grep MaxAlloc
      MaxAlloc: 4.194304e+08

helm template charts/refinery --set config.InMemCollector.MaxAlloc=42 | grep MaxAlloc
      MaxAlloc: 42
```

## Potential Drawbacks

Due to the minimal nature of math functions in Helm, we don't currently handle string values for memory resources, only resources denoted in bytes as an integer.

### DIFF
--- a/charts/refinery/templates/configmap-config.yaml
+++ b/charts/refinery/templates/configmap-config.yaml
@@ -6,4 +6,36 @@ metadata:
   {{- include "refinery.labels" . | nindent 4 }}
 data:
   config.yaml: |
-{{- toYaml .Values.config | nindent 4 }}
+    EnvironmentCacheTTL: {{ .Values.config.EnvironmentCacheTTL }}
+    GRPCListenAddr: {{ .Values.config.GRPCListenAddr }}
+    HoneycombAPI: {{ .Values.config.HoneycombAPI }}
+    InMemCollector:
+      CacheCapacity: {{ .Values.config.InMemCollector.CacheCapacity }}
+
+      {{- if hasKey .Values.config.InMemCollector "MaxAlloc" }}
+      MaxAlloc: {{ .Values.config.InMemCollector.MaxAlloc }}
+      {{ else }}
+      MaxAlloc: {{ mulf .Values.resources.requests.memory 0.8 }}
+      {{- end }}
+
+    ListenAddr: {{ .Values.config.ListenAddr }}
+    Logger: {{ .Values.config.Logger }}
+    LoggingLevel: {{ .Values.config.LoggingLevel }}
+    MaxBatchSize: {{ .Values.config.MaxBatchSize }}
+    Metrics: {{ .Values.config.Metrics }}
+    PeerBufferSize: {{ .Values.config.PeerBufferSize }}
+    PeerListenAddr: {{ .Values.config.PeerListenAddr }}
+    PeerManagement:
+      IdentifierInterfaceName: {{ .Values.config.PeerManagement.IdentifierInterfaceName }}
+      RedisHost: {{ .Values.config.PeerManagement.RedisHost }}
+      RedisPassword: "{{ .Values.config.PeerManagement.RedisPassword }}"
+      Type: {{ .Values.config.PeerManagement.Type }}
+      UseIPV6Identifier: {{ .Values.config.PeerManagement.UseIPV6Identifier }}
+      UseTLS: {{ .Values.config.PeerManagement.UseTLS }}
+    PrometheusMetrics:
+      MetricsListenAddr: {{ .Values.config.PrometheusMetrics.MetricsListenAddr }}
+    RulesConfigMapName: "{{ .Values.config.RulesConfigMapName }}"
+    SendDelay: {{ .Values.config.SendDelay }}
+    SendTicker: {{ .Values.config.SendTicker }}
+    TraceTimeout: {{ .Values.config.TraceTimeout }}
+    UpstreamBufferSize: {{ .Values.config.UpstreamBufferSize }}
--- a/charts/refinery/templates/configmap-config.yaml
+++ b/charts/refinery/templates/configmap-config.yaml
@@ -15,7 +15,7 @@ data:
       {{- if hasKey .Values.config.InMemCollector "MaxAlloc" }}
       MaxAlloc: {{ .Values.config.InMemCollector.MaxAlloc }}
       {{ else }}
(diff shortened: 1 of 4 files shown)


charts/refinery/values.yaml
```diff
@@ -19,10 +19,10 @@ resources:
   limits:
     cpu: 2000m
     # Changing memory limits requires updating config.InMemCollector.MaxAlloc
-    memory: 2Gi
+    memory: 2097152000 # use bytes to allow auto-configuration of `MaxAlloc`
```
MikeGoldsmith, line 22: 2GB is 2147483648 bytes

```suggestion
    memory: 2147483648 # use bytes to allow auto-configuration of `MaxAlloc`
```

charts/refinery/templates/configmap-config.yaml
```diff
@@ -6,4 +6,36 @@ metadata:
...
+    EnvironmentCacheTTL: {{ .Values.config.EnvironmentCacheTTL }}
+    GRPCListenAddr: {{ .Values.config.GRPCListenAddr }}
+    HoneycombAPI: {{ .Values.config.HoneycombAPI }}
+    InMemCollector:
+      CacheCapacity: {{ .Values.config.InMemCollector.CacheCapacity }}
+
+      {{- if hasKey .Values.config.InMemCollector "MaxAlloc" }}
+      MaxAlloc: {{ .Values.config.InMemCollector.MaxAlloc }}
+      {{ else }}
+      MaxAlloc: {{ mulf .Values.resources.requests.memory 0.8 }}
```
MikeGoldsmith, line 18: We say 75% in README but are multiplying by 0.8 (80%).

```suggestion
      MaxAlloc: {{ mulf .Values.resources.requests.memory 0.75 }}
```

charts/refinery/values.yaml
```diff
@@ -176,7 +176,7 @@ config:
     # This value should be set in according to the resources.limits.memory
     # By default that setting is 2GB, and this is set to 85% of that limit
     # 2 * 1024 * 1024 * 1024 * 0.80 = 1,717,986,918
-    MaxAlloc: 1717986918
+    # MaxAlloc: 42
```
MikeGoldsmith, line 179: I don't think we need to update this value

```suggestion
    # MaxAlloc: 1717986918
```

MikeGoldsmith (changes requested):
This looks really good, thanks @bixu 👍🏻 

I've left some minor feedback for consistency.

charts/refinery/values.yaml
bixu: Yeah, this is another place where I was confused about the difference between Gigabytes and Gibibytes. Thanks for correcting.

charts/refinery/templates/configmap-config.yaml
bixu: Yes, good catch. I actually wanted to ask about this -- I was assuming the more conservative value is best, but happy to change the behavior if needed.

charts/refinery/values.yaml
bixu: Agree. This was a little joke for Hitchhiker's fans that I intended to pull after exiting draft.

bixu:
I'm in the middle of a role change at $job so pls ping me on Pollinators if there are more questions or this is accepted -- a bit of a 🤹🏼 atm :D

charts/refinery/templates/configmap-config.yaml
```diff
@@ -6,4 +6,36 @@ metadata:
   {{- include "refinery.labels" . | nindent 4 }}
 data:
   config.yaml: |
-{{- toYaml .Values.config | nindent 4 }}
+    EnvironmentCacheTTL: {{ .Values.config.EnvironmentCacheTTL }}
```
MikeGoldsmith, line 9: Missing properties not mapped from values yaml:
- BatchTimeout
- AdditionalErrorFields
- HoneycombLogger (with sub properties)
- HoneycombMetrics (with sub properties)

MikeGoldsmith (changes requested):
Looks like we're missing some properties being copied over from values yaml.

@kentquirk @TylerHelmuth could you also take a look too please?

charts/refinery/templates/configmap-config.yaml
TylerHelmuth: I'm new to this repo, but I'm a little concerned about listing out each `.Values.config` option in the config map.  If we add new config options in the future then we have to add them both to the values.yaml and in this file.  Instead, can we use some tpl helpers to handle the new `MaxAlloc` logic?  We should be able to check if the field is set in config using `get`.  [We do something similar in the opentelemetry-helm-charts.](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/fe8ee564b0c6c76f14e3af20daf318ae28855dce/charts/opentelemetry-collector/templates/_config.tpl#L26-L35)

bixu: @TylerHelmuth, I had the same concern, but didn't know already of a good way to do this with template helpers. Thanks for the example -- will take a crack at that.

TylerHelmuth: @bixu no problem.  Inside the [opentelemetry-helm-charts](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/fe8ee564b0c6c76f14e3af20daf318ae28855dce/charts/opentelemetry-collector/templates/_config.tpl) there are some helper templates that do some calculating.  The `"opentelemetry-collector.convertMemToMib"` definition and `"opentelemetry-collector.getMemLimitMib"` will hopefully be helpful.

TylerHelmuth: Sorry @bixu I might have misunderstood your question.  You commit looks like it is doing a calculation if `.Values.config.InMemCollector "MaxAlloc"` is not set.  Is the calculation not working as expected?

PR #104, opened 2022-01-27
Allow operators to disable live reload of Refinery related to changed sampling rules

bixu:
This PR is a workaround for the issue describe in #101

PR #189, opened 2022-10-13
fix: Use distinct annotation keys for checksums

mikebryant:

Currently both checksums use the same key, so only one is present in the resulting Deployment. This is leading to our Deployments not restarting on rule changes.

Change the key for the rules so it's distinct from the one for config.

Deploy the chart, observe that there are now 2 checksum annotations on the Pod instead of only one.

### DIFF
--- a/charts/refinery/templates/deployment.yaml
+++ b/charts/refinery/templates/deployment.yaml
@@ -24,7 +24,7 @@ spec:
         prometheus.io/scrape: "true"
         {{- end }}
         {{- if hasKey .Values.rules "LiveReload" | ternary (not .Values.rules.LiveReload) false }}
-        checksum/config: {{ include (print $.Template.BasePath "/configmap-rules.yaml") . | sha256sum }}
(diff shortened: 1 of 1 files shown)


TylerHelmuth (approved):
Great catch

TylerHelmuth (commented):
@mikebryant I was over zealous with my initial approval and thinking about things more I think this is a proper change, but I want to make sure we've covered all the test scenarios.

At the moment if `.Values.rules.LiveReload` is set to false then the deployment ends up with two `checksum/config` entries, which is definitely not proper and should be fixed.

I believe the original PR's intent was to allow disabling of refinery's LiveReload for rules: https://github.com/honeycombio/helm-charts/pull/104

Can you test and ensure that 
- If `.Values.rules.LiveReload` is set to false, changing the config leads to deployments restarting (existing behavior)?
- If `.Values.rules.LiveReload` is set to false, changing the rules leads to deployments restarting (new behavior)?

MikeGoldsmith:
Closing in favour of #197

PR #193, opened 2022-10-18
[chore] add chart testing

TylerHelmuth:

Currently all tests of the helm charts have to be done manually.  This can be time consuming and it can be difficult to cover all useful test scenarios

Adds a new circleci workflow to use [chart-testing](https://github.com/helm/chart-testing) to test the helm charts. This allows automatic detection of changed charts and then installs all the values.yamls in the `ci` folder, one at a time, to ensure the chart is installable.

copious amounts of trial and error (118 to be exact) lol


masonjlegere:
Nice! Probably unrelated as this seems pretty e2e but I've used [this plugin](https://github.com/quintush/helm-unittest) for setting up fast unit tests without having to setup minikube/k3d etc

TylerHelmuth:
@masonjlegere thanks for the suggestion, I've not heard of that one before.  I've always used chart-testing to test install helm charts.  Annoyingly it does require minikube and kubectl and helm, but I like that it actually does an install of all the scenarios in the ci folder.

This PR would be useful for the PR you linked, but it'll be useful for all PRs going forward as well, especially once we flush out the test scenarios.

charts/refinery/values.yaml
```diff
@@ -164,7 +164,7 @@ config:
     # Further, if the environment variable 'REFINERY_REDIS_HOST' is set it takes
     # precedence and this value is ignored.
     # Not eligible for live reload.
-    RedisHost: "refinery-redis:6379"
+    RedisHost: '{{ include "refinery.redis.fullname" . }}:6379'
```
TylerHelmuth, line 167: leaving this PR in draft until https://github.com/honeycombio/helm-charts/pull/194 is decided.

MikeGoldsmith (changes requested):
Looks good - only suggestion would be to make the release step require the test step to ensure it doesn't continue if it fails.

 You can add the following to the `release-charts` workflow
```yaml
          requires:
            - test
```

TylerHelmuth:
@MikeGoldsmith it will be good to run the tests one more time before releasing, but we'll need to think about how to do this.  I don't think adding the `test` job I wrote will work in this instance since it is taking advantage of chart-testing's ability to detect changes between the PR's chart and the main branch's chart.  For the release workflow, it won't find any changes and won't run any tests.

I think we should still move forward with this PR to get automated tests on our PRs while we continue to iterate on how to best test before the release is pushed.

robbkidd:
> chart-testing […] detect[s] changes between the PR's chart and the main branch's chart

Would "test-modified-charts" be a better name for this job? That could manage the reader's expectations and be less surprising that the job doesn't test anything if there are no changes to a chart in a PR.

The "test" job in the majority of our other projects involves testing everything, modified and unmodified.

robbkidd (approved):
Makes sense to me. MOAR TESTING

a man wearing a helmet hits himself repeatedly with a stick

MikeGoldsmith:
> @MikeGoldsmith it will be good to run the tests one more time before releasing, but we'll need to think about how to do this. I don't think adding the `test` job I wrote will work in this instance since it is taking advantage of chart-testing's ability to detect changes between the PR's chart and the main branch's chart. For the release workflow, it won't find any changes and won't run any tests.
> 
> I think we should still move forward with this PR to get automated tests on our PRs while we continue to iterate on how to best test before the release is pushed.

Thanks for the clarification on how the tests are expected to run and I agree it doesn't makes sense to make the test step a prerequisite for releasing here 👍🏻

MikeGoldsmith (approved):

PR #194, opened 2022-10-18
[refinery] Update to set redis host based on release name

TylerHelmuth:

Currently the refinery chart must be installed with a release name of `refinery` or the `.Values.config.PeerManagement.RedisHost` must be manually updated to match the release name.  This works if you know what the release name will be, but for situations like chart-testing, the release name is dynamically generated.  It is also problematic when using the chart for the first time; if you want to use the default values.yaml you must use `refinery` as the release name.

Updates `.Values.config` to be run through the `tpl` function.  This allows us to set `RedisHost: '{{include "refinery.redis.fullname" .}}:6379'` in the values.yaml and us the chart's fullname helpers to determine the default RedisHost.

Deployed refinery local and ran it through chart-testing.

### DIFF
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -164,7 +164,7 @@ config:
     # Further, if the environment variable 'REFINERY_REDIS_HOST' is set it takes
     # precedence and this value is ignored.
     # Not eligible for live reload.
-    RedisHost: "refinery-redis:6379"
+    RedisHost: '{{include "refinery.redis.fullname" .}}:6379'
 
     # RedisUsername is the username used to connect to redis for peer cluster membership management.
     # If the environment variable 'REFINERY_REDIS_USERNAME' is set it takes
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -66,6 +66,8 @@ extraVolumes: [ ]
 
 # Values used to build config.yaml.
 # See the example in sample-configs/config_complete.yaml for full details on all properties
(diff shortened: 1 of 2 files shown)


JamieDanielson (commented):
To confirm - in the previous scenario, the name was hard-coded to "refinery", and would have required a change by the user to use something other than the hard-coded name of "refinery".
With this change, for existing users it should use the already-set value of "refinery" or the already-set custom value replacing "refinery". For new users, it will use whatever value they choose (so they are no longer required to use "refinery") or whatever value is generated (as in the case of testing or other auto-generated scenarios).

I'm trying to double check whether this would break existing behavior for anyone, or result in unexpected behavior by using a name other than the initially hard-coded "refinery".

charts/refinery/values.yaml
```diff
@@ -66,6 +66,8 @@ extraVolumes: [ ]
 
 # Values used to build config.yaml.
 # See the example in sample-configs/config_complete.yaml for full details on all properties
+# Supports templating. To escape existing instances of {{ }}, use {{` <original content> `}}.
+# For example, {{ REDACTED_EMAIL }} becomes {{` {{ REDACTED }} `}}.
```
JamieDanielson, line 70:
```suggestion
# For example, {{ REDACTED_EMAIL }} becomes {{` {{ REDACTED_EMAIL }} `}}.
```

I think the example of escaping is just adding backticks, right?

TylerHelmuth: Extra {{ }} as well, but the rename was unintended.

JamieDanielson (approved):
Awesome, this is a handy update. Thanks!

Issue #191, opened 2022-10-13
Add Automated Tests for all charts


TylerHelmuth:
**Is your feature request related to a problem? Please describe.**

Currently there is no automated test execution for PRs.  As a result, we must rely on manual tests to ensure changes are accurate.

- [x] add new workflow
- [ ] add CI scenarios to honeycomb
- [x] add CI scenarios to opentelemetry-collector
- [x] add CI scenarios to refinery

**Describe the solution you'd like**

Add a circleci workflow that uses [chart-testing](https://github.com/helm/chart-testing) to test chart scenarios.

TylerHelmuth:
https://github.com/honeycombio/helm-charts/pull/193

TylerHelmuth:
I don't think Honeycomb chart has any unique scenarios outside of the default, so calling that done for now

PR #195, opened 2022-10-24
[refinery] Add more refinery scenarios

TylerHelmuth:
A default test scenario was recently added to the refinery chart, but it does not cover all scenarios.

- Works toward #191

Adds more test scenarios for the refinery chart, targeting situations where decisions are made when rendering based on the values.yaml

Automated tests


TylerHelmuth:
Removed the redis scenario since we don't have an actual existing redis instance to point to so that refinery can actually start.

MikeGoldsmith (approved):

Issue #216, opened 2023-01-19
Reorder the Resource metrics to have memory before CPU 


robiball:
**Is your feature request related to a problem? Please describe.**
HPA objects incorrectly show as `out of sync` in ArgoCD if the metrics ordering doesn't match what's expected. This has been resolved in the `opentelemetry-collector` chart (see here: https://github.com/honeycombio/helm-charts/pull/164), but is still present in the `refinery` and `secure-tenancy` charts.

**Describe the solution you'd like**
Reorder the resouce metrics in the `refinery` and `secure-tenancy` charts to have memory before cpu.

**Additional context**
Happy to PR this in if you agree with it.

TylerHelmuth:
@robiball thanks for the details.  Would love a PR to fix this.

robiball:
https://github.com/honeycombio/helm-charts/pull/217 :)

PR #217, opened 2023-01-19
fix: Updated ordering of CPU/Memory resource metrics

robiball:

- Closes https://github.com/honeycombio/helm-charts/issues/216

Changes the ordering of the resouce metrics in the `refinery` and `secure-tenancy` charts to have memory before cpu. This ensures consistency with the HPA in the `opentelemetry-collector` chart, and stops ArgoCD from thinking these HPAs are always out of sync.

### DIFF
--- a/charts/refinery/templates/hpa.yaml
+++ b/charts/refinery/templates/hpa.yaml
@@ -14,14 +14,6 @@ spec:
   minReplicas: {{ .Values.autoscaling.minReplicas }}
   maxReplicas: {{ .Values.autoscaling.maxReplicas }}
   metrics:
-    {{- if .Values.autoscaling.targetCPUUtilizationPercentage }}
-    - type: Resource
-      resource:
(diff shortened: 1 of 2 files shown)


charts/refinery/templates/hpa.yaml
```diff
@@ -15,20 +15,20 @@ spec:
   maxReplicas: {{ .Values.autoscaling.maxReplicas }}
   metrics:
     {{- if .Values.autoscaling.targetCPUUtilizationPercentage }}
+    {{- if .Values.autoscaling.targetMemoryUtilizationPercentage }}
```
TylerHelmuth, line 18: Looks like the CPU if statement needs moved.

robiball: Good spot, I've pushed a fix.

TylerHelmuth (approved):

PR #220, opened 2023-01-26
docs: Fix the unit for metrics.interval

DavidS-om:

setting metrics.interval to 300 leads to this startup log:
```
time="2023-01-26T16:26:25Z" level=info msg="Creating Metrics Service Runner..."  [...] interval=300ns
```

and LOTS of events sent to honeycomb


puckpuck:
The field is a `time.Duration` field. So you can pass in things like `300s` to it. We should update the table to indicate this instead of calling it a nanosecond field.

TylerHelmuth:
@DavidS-om are you able to update the table based on @puckpuck's feedback?

DavidS-om:
@TylerHelmuth I love honeycomb dearly, but I've fixed my problem and I do not know (nor want to know) what a `time.Duration` field is or can do.

TylerHelmuth:
@DavidS-om thank you.  I am going to close this PR in favor of https://github.com/honeycombio/helm-charts/pull/221

PR #222, opened 2023-02-09
chore: add nodeport to the service.yaml

fchikwekwe:
Port mapping is useful for testing refinery locally. Adding this to the `service.yaml` so that it does not need to manually be added by anyone needing to reference this helm chart.


kentquirk (approved):
Thanks!

PR #227, opened 2023-03-13
maint: Update refinery chart to use refinery 1.20.0

TylerHelmuth:

- Closes https://github.com/honeycombio/helm-charts/issues/223

- Updates refinery chart to use refinery 1.20.0 by default.  Updated values.yaml with new configuration fields.  Updated `NOTES.txt` to log [INFO] when values are not set to recommended values.

Local kind cluster.

### DIFF
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -167,6 +167,24 @@ config:
   # Eligible for live reload.
   CacheOverrunStrategy: "impact"
 
+  # Set the field names to use for the event ID fields. These fields are used to
+  # identify events that are part of the same trace. TraceIdFieldNames are used
+  # to determine if an event is a Trace or another event type. ParentIdFieldNames
+  # are used to determine if an event is a root span or not.
+  # TraceIdFieldNames:
+  #   - "trace.trace_id"
+  #   - "traceId"
+  # ParentIdFieldNames:
+  #   - "trace.parent_id"
+  #   - "parentId"
+
+  # AdditionalAttributes is a map that can be used for injecting user-defined
+  # attributes. For example, it could be used for naming a refinery cluster.
+  # Both keys and values must be strings.
+  # AdditionalAttributes:
+    # ClusterName: MyCluster
+    # environment: production
+
   # Configure how Refinery peers are discovered and managed
   PeerManagement:
     # The type should always be redis when deployed to Kubernetes environments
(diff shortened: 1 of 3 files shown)


charts/refinery/values.yaml
```diff
@@ -206,6 +236,15 @@ config:
...
     UseIPV6Identifier: false
 
+    # Strategy controls the way that traces are assigned to refinery nodes.
+    # The "legacy" strategy uses a simple algorithm that unfortunately causes
+    # 1/2 of the in-flight traces to be assigned to a different node whenever the
+    # number of nodes changes.
+    # The legacy strategy is deprecated and is intended to be removed in a future release.
+    # The "hash" strategy is strongly recommended, as only 1/N traces (where N is the
+    # number of nodes) are disrupted when the node count changes.
+    Strategy: "legacy"
```
kentquirk, line 246: Given that you're adding this stanza for the first time, might it make sense to just set it to "hash"?

charts/refinery/values.yaml
```diff
@@ -228,6 +267,96 @@ config:
...
+  # status after the sampling decision has been made.
+  SampleCacheConfig:
+
+    # Type controls the type of sample cache used.
+    # "legacy" is a strategy where both keep and drop decisions are stored in a circular buffer that is
+    # 5x the size of the trace cache. This is Refinery's original sample cache strategy.
+    # "cuckoo" is a strategy where dropped traces are preserved in a "Cuckoo Filter", which can remember
+    # a much larger number of dropped traces, leaving capacity to retain a much larger number of kept traces.
+    # It is also more configurable. The cuckoo filter is recommended for most installations.
+    Type: "legacy"
```
kentquirk, line 339: similarly here -- could we set this to cuckoo in the chart?

kentquirk (approved):
I have questions about defaults, but I'm happy either way.

charts/refinery/values.yaml
TylerHelmuth: I think it would be considered a breaking change for the same reason we didn't default refinery to hash. If we set the default value to hash in the chart and someone takes an upgrade without setting it back to legacy then they just started using hash without opting in.

charts/refinery/values.yaml
TylerHelmuth: Same reason as before. Since we didn't use it as a default value in refinery then the same rule would apply here. Defaulting to cuckoo here is the same as defaulting to cuckoo in refinery.

PR #240, opened 2023-04-20
rel: prep refinery 1.19.0 release

TylerHelmuth:

- prep release of Refinery chart

- bump chart version
- update changelog

### DIFF
--- a/charts/refinery/Chart.yaml
+++ b/charts/refinery/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 1.18.0
+version: 1.19.0
 appVersion: 1.21.0
 keywords:
   - refinery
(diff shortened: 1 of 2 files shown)


charts/refinery/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 1.18.0
+version: 1.19.0
```
akvanhar, line 5: Newb question: It's interesting to me that the version of the chart is different than the version of refinery. Do we ever try to keep those the same?

akvanhar (approved):

charts/refinery/Chart.yaml
TylerHelmuth: A helm chart has an independent release and version for the thing it manages.  The main reason for this is because the chart might have features/bug fixes/breaking changes that are independent of the underlying service that need to be release to the helm chart users.  For example, if we needed to make a change to the chart to handle some new k8s feature, we could release that without needing to release Refinery.

PR #242, opened 2023-04-26
feat: Added support for additional env variable settings

### LINKED ISSUE #241: Setting specific proxy setting for a kubernetes deployment
We need to set specific proxy settings in a kubernetes deployment. Currently there is no way to set this via the values file or env variables.  You can see like this is handled at New-Relic: [configure-kubernetes-proxy](https://docs.newrelic.com/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/configure-kubernetes-proxy/) 

I've implemented a solution which would not only add the proxy settings, but also allow to set additional env variables.
You can find the solution here: [https://github.com/honeycombio/helm-charts/compare/main...nexineer:helm-charts:main](https://github.com/honeycombio/helm-charts/compare/main...nexineer:helm-charts:main)

anux-linux:

Adding additional custom env variables to honeycomb

- Closes #241

extending config to read env variable from a user values file.

adding proxy setting like this will add them to the chart
```yaml
additionalEnv:
  - name: HTTP_PROXY
    value: http://your-proxy-url.example:8080
```

### DIFF
--- a/charts/honeycomb/README.md
+++ b/charts/honeycomb/README.md
@@ -38,6 +38,10 @@ watchers:
     labelSelector: component=kube-scheduler
     namespace: kube-system
     parser: glog
+# Optional your additional env settings
+additionalEnv:
+  - name: HTTP_PROXY
+    value: http://your-proxy-url.example:8080
 ```
 Then use this yaml file when installing the chart
 ```bash
--- a/charts/honeycomb/README.md
+++ b/charts/honeycomb/README.md
@@ -38,10 +38,6 @@ watchers:
     labelSelector: component=kube-scheduler
     namespace: kube-system
     parser: glog
-# Optional your additional env settings
-additionalEnv:
-  - name: HTTP_PROXY
-    value: http://your-proxy-url.example:8080
 ```
 Then use this yaml file when installing the chart
 ```bash
--- a/charts/honeycomb/templates/daemonset.yaml
+++ b/charts/honeycomb/templates/daemonset.yaml
@@ -62,8 +62,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.hostIP
-            {{- if gt (len .Values.additionalEnv) 0 -}}
-              {{- toYaml .Values.additionalEnv | nindent 12}}
+            {{- with .Values.additionalEnv -}}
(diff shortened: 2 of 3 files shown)


charts/honeycomb/templates/daemonset.yaml
```diff
@@ -62,6 +62,9 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.hostIP
+            {{- if gt (len .Values.additionalEnv) 0 -}}
+              {{- toYaml .Values.additionalEnv | nindent 12}}
+            {{- end }}
```
TylerHelmuth, line 67: Instead of a condition we can use `with`.  Something like

```yaml
{{- with .Values.additionalEnv }}
  {{- . | toYaml | nindent 12 }}
{{- end }}
```

anux-linux: Thanks for the info. wasn't aware of this notion. I updated the code to use `with`

charts/honeycomb/README.md
```diff
@@ -38,6 +38,10 @@ watchers:
     labelSelector: component=kube-scheduler
     namespace: kube-system
     parser: glog
+# Optional your additional env settings
+additionalEnv:
+  - name: HTTP_PROXY
+    value: http://your-proxy-url.example:8080
```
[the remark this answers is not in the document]
anux-linux: I removed the lines from README and added a comment to the values.yaml file.

charts/honeycomb/values.yaml
```diff
@@ -97,6 +97,11 @@ resources: {}
   #   cpu: 100m
   #   memory: 128Mi
 
+# You can add additional environment values like this:
+#additionalEnv:
```
TylerHelmuth, line 101:
```suggestion
additionalEnv: {}
```

charts/honeycomb/values.yaml
```diff
@@ -97,6 +97,11 @@ resources: {}
   #   cpu: 100m
   #   memory: 128Mi
 
+# You can add additional environment values like this:
```
TylerHelmuth, line 100:
```suggestion
# Field for adding additional environment variables
```

TylerHelmuth (approved):

PR #243, opened 2023-05-02
feat: Add the ability to create your own Secret Provider

RainofTerra:

This allows you to create your own secretProvider.

Adds a secretprovider.yaml and adds example values to the values.yaml

uncomment out the example and replace the environment name with the correct one.

### DIFF
--- a/charts/refinery/templates/secretprovider.yaml
+++ b/charts/refinery/templates/secretprovider.yaml
@@ -1,0 +1,11 @@
+{{- if .Values.secretProvider.create -}}
+apiVersion: secrets-store.csi.x-k8s.io/v1
+kind: SecretProviderClass
+metadata:
+  name: {{ .Values.secretProvider.name }}
+  namespace: {{ .Release.Namespace }}
+spec:
+  {{- with .Values.secretProvider.spec }}
+  {{- toYaml . | nindent 4 }}
+  {{- end }}
+{{- end }}
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -565,3 +565,25 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+
+
+# If you need to create a secret provider, such as for using AWS SSM, you can do so here:
+#
+# secretProvider:
+#   create: true
+#   spec:
+#     provider: aws
+#     secretObjects:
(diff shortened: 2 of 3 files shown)


charts/refinery/templates/secretprovider.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.secretProvider.create -}}
+apiVersion: secrets-store.csi.x-k8s.io/v1
+kind: SecretProviderClass
```
TylerHelmuth, line 3: I'm not familiar with this type of resource.  Can it be installed into any k8s cluster?

RainofTerra: More information is available [here](https://secrets-store-csi-driver.sigs.k8s.io/), but yes I think any modern (>1.19) cluster should be able to support it.

charts/refinery/values.yaml
```diff
@@ -565,3 +565,25 @@ nodeSelector: {}
...
+#         - key: yourenvironment.refinery_honeycomb_api_key
+#           objectName: youenvironment.refinery_honeycomb_api_key
+#     parameters:
+#       objects: |
+#           - objectName: yourenvironment.refinery_honeycomb_api_key
+#             objectType: "ssmparameter"
+#   name: "refinery"
+
+secretProvider:
+  create: false
```
TylerHelmuth, line 589: Can you add `name: refinery` and `spec: {}` as default values to this section.

RainofTerra: Done!

charts/refinery/templates/secretprovider.yaml
TylerHelmuth: It looks like it does require a CustomResourceDefinition.  Normally a chart should but installable without external dependencies, but since this is optional I think it is ok.  Can you add in the values.yaml a comment linking to those docs and describing that the chart will only be installable with this feature enabled if those CRDs are present?

TylerHelmuth: Also in the readme table.

RainofTerra: Done!
image

TylerHelmuth (approved):

PR #244, opened 2023-05-04
rel: prep Refinery 1.19.1 release

TylerHelmuth:

Prepare refinery chart 1.19.1 release

- Bump chart version
- update changelog

### DIFF
--- a/charts/refinery/CHANGELOG.md
+++ b/charts/refinery/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Refinery Helm Chart Changelog
 
+## Refinery v1.19.1
+  - feat: Add the ability to create your own Secret Provider (#243) | @RainofTerra
+  - chore: Allow users to add annotations for refinery and collector deployments (#236) | @suryapandian
+
 ## Refinery v1.19.0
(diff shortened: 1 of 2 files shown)


charts/refinery/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Refinery Helm Chart Changelog
 
+## Refinery v1.19.1
+  - feat: Add the ability to create your own Secret Provider (#243) | @RainofTerra
```
kentquirk, line 4: It doesn't look like we've done it the last couple of times, but in the past and in other repositories, we use a link to the user instead of just the @. Example: `[@TylerHelmuth](https://github.com/TylerHelmuth)`

kentquirk (approved):
I feel like it would be good to fix the user links, but I won't block this if you want to do it at a different time.

charts/refinery/CHANGELOG.md
TylerHelmuth: Ya I've been purposefully updating the user links bc the `@`s in the release notes auto-generate links and more importantly auto-generate the Contributors section.

PR #245, opened 2023-05-11
maint: deprecate secure-tenancy chart

TylerHelmuth:
Prepares the secure-tenancy chart for deprecation.

Following https://helm.sh/docs/topics/charts/#deprecating-charts and https://github.com/helm/charts/blob/master/PROCESSES.md#deprecating-a-chart, from this PR I will release the chart one last time and then submit a follow-up PR to delete it.


kentquirk (approved):
Boom.

PR #249, opened 2023-05-12
maint: Remove secure tenancy chart

TylerHelmuth:

Removes the deprecated secure tenancy chart and all references to it.

- removes secure-tenency chart
- update main README
- remove unneeded dependencies from install script


kentquirk (approved):
Red diffs are the best diffs.

PR #250, opened 2023-05-12
maint: cleanup release

TylerHelmuth:

With the bump to chart releaser 1.5.0 we lost the ability to do `cr index` with an `ssh` git remote; the tool now requires us to use an `https` remote.  For the `v1.2.3-secure-tenancy` release I had to do the `cr index` step manually from my laptop. See and https://github.com/helm/chart-releaser/issues/124.

The PR solves this problem and also removes some deprecated flags and unused env vars.

- set git remote to https url for `cr index` step.
- remove unused `--charts-repo` flag
- remove unused variables.

I couldn't :/ but I'll try to do a honeycomb chart release once this is merged to test it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,6 @@ jobs:
       - run:
           command: |
             echo "export GIT_USERNAME=$CIRCLE_USERNAME" >> $BASH_ENV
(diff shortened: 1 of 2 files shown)


kentquirk (approved):
LGTM.

PR #254, opened 2023-06-23
docs: [refinery] Add MaxConnectionAge default to values.yaml

mterhar:

The MaxConnectionAge default is 0 which leads to load balancers failing to re-balance GRPC connections to Refinery. This means that one node in a multi-node deployment receives all of the traffic and expends more resources than it should redirecting all the traffic while the other nodes sit idle. 

This PR doesn't change the default but exposes it as a configuration that can be set so operators can be in a better position to fix the problem if they run into it. 

Adds some new  

You can use the `helm template` command to see that the config is present. When it comes to actually verifying the connections remain open, I haven't cracked that nut yet.

### DIFF
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -267,6 +267,19 @@ config:
     # By default that setting is 2GB, and this is set to 85% of that limit
     # 2 * 1024 * 1024 * 1024 * 0.80 = 1,717,986,918
     MaxAlloc: 1717986918
+    
+  # Reflects: https://pkg.go.dev/google.golang.org/grpc/keepalive#ServerParameters
+  GRPCServerParameters:
+  
+    # MaxConnectionAge is a duration for the maximum amount of time a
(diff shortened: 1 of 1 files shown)


TylerHelmuth:
> This PR doesn't change the default but exposes it as a configuration that can be set so operators can be in a better position to fix the problem if they run into it.

To be clear, this PR is not necessary for the chart to be able to set this value - helm chart users are allowed to set this value in the values.yaml today.  I see the value of this PR as showing users who read the default values.yaml that this is a configuration that Refinery allows.

TylerHelmuth (approved):
I am good with merging this so that people can read main's `values.yaml` and see this field, but with Refinery 2.0 releasing soon<sup>TM</sup> we don't plan to do another release of the chart until the 2.0 release.

mterhar:
@TylerHelmuth Thanks for the feedback. I agree that it's just to help with future reference for refinery 1.x users. 

Can you merge? I lack permission.

PR #255, opened 2023-06-26
feat: enable debug service by default

lizthegrey:

pprof endpoints are not enabled by default and cannot easily be enabled since the argv isn't editable from the variables passed to the helm template by users.

Enables the -d flag by default, which cause the container to listen on the :6060 debug service name.

### DIFF
--- a/charts/refinery/ci/debug.yaml
+++ b/charts/refinery/ci/debug.yaml
@@ -1,0 +1,11 @@
+resources:
+  limits:
+    cpu: 100m
+    memory: 200M
+  requests:
+    cpu: 100m
+    memory: 200M
+
+debug:
+  enabled: true
+  port: 6062
--- a/charts/refinery/templates/deployment.yaml
+++ b/charts/refinery/templates/deployment.yaml
@@ -50,6 +50,7 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
             - "refinery"
+            - "-d"
             - "-c"
             - "/etc/refinery/config.yaml"
             - "-r"
@@ -59,6 +60,9 @@ spec:
           {{- toYaml . | nindent 12 }}
           {{- end }}
           ports:
+            - name: debug
+              containerPort: 6060
+              protocol: TCP
             - name: data
               containerPort: 8080
               protocol: TCP
--- a/charts/refinery/templates/deployment.yaml
+++ b/charts/refinery/templates/deployment.yaml
@@ -50,19 +50,18 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
             - "refinery"
-            - "-d"
             - "-c"
             - "/etc/refinery/config.yaml"
             - "-r"
             - "/etc/refinery/rules.yaml"
+            {{- if .Values.debug.enabled }}
+            - "-d"
+            {{- end }}
           {{- with .Values.environment }}
           env:
           {{- toYaml . | nindent 12 }}
           {{- end }}
           ports:
-            - name: debug
-              containerPort: 6060
-              protocol: TCP
             - name: data
               containerPort: 8080
               protocol: TCP
@@ -75,6 +74,11 @@ spec:
             {{- if eq .Values.config.Metrics "prometheus" }}
             - name: metrics
               containerPort: 9090
+              protocol: TCP
+            {{- end }}
+            {{- if .Values.debug.enabled }}
+            - name: debug
+              containerPort: {{ .Values.debug.port }}
               protocol: TCP
             {{- end }}
           volumeMounts:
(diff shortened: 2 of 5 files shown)


charts/refinery/templates/deployment.yaml
```diff
@@ -50,6 +50,7 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           command:
             - "refinery"
+            - "-d"
```
TylerHelmuth, line 53: Is there a performance cost to enabling this by default?  Instead of enabling this by default I think we could define a new root-level, debug-specific field that when enabled would set this flag and the necessary port.

lizthegrey: No cost to doing it (unless something is pulling pprof continuously and even then <1% cpu impact); it's purely a security thing, but if people can poke arbitrary ports on your k8s cluster's pods without using the ingress you have other things severely wrong.

charts/refinery/templates/deployment.yaml
```diff
@@ -59,6 +60,9 @@ spec:
           {{- toYaml . | nindent 12 }}
           {{- end }}
           ports:
+            - name: debug
+              containerPort: 6060
```
TylerHelmuth, line 64: Refinery's default behavior is to use the first available port between :6060 and :6069 when `DebugServiceAddr` is not set.  Since the helm chart doesn't set this value by default, we cannot rely on this value always being the port chosen by Refinery.

charts/refinery/templates/deployment.yaml
TylerHelmuth: Maybe it is because of how we've named to capability, but it does not feel proper to turn on debug mode by default.  This mode enables profiling, but also enables some more verbose logging, at least during startup.

I think I'd still prefer a new `debug` section in the values.yaml to configure this value and the port.   Something like

```yaml
debug:
  enabled: false
  port: 6060
```

We can introduce some logic to set `DebugServiceAddr` in the config if `debug.port` is set and `DebugServiceAddr` is not set and/or fail the install via NOTES.txt if  `debug.port` does not equal `DebugServiceAddr`.

charts/refinery/templates/deployment.yaml
lizthegrey: If it's the only thing running in the container, why wouldn't it be able to bind :6060? oh, unless the config is explicitly setting DebugServiceAddr to something else, you're suggesting?

charts/refinery/templates/deployment.yaml
lizthegrey: I don't understand Helm enough to do this, but if you want to modify my PR to do this feel free :)

charts/refinery/templates/deployment.yaml
TylerHelmuth: Correct. If the user explicitly sets the value hard coding this definitely won't work. If they have chosen to bind something else to port 6060 (via some config value) this assumption also breaks down.

charts/refinery/templates/deployment.yaml
TylerHelmuth: With Refinery 2.0 release, Q3 planning, and hack week I don't anticipate being able to implement the changes myself this week. What is the priority of this work?

lizthegrey: System optimization work I was hoping to have for https://github.com/honeycombio/infra/pull/4432 - which I'm hoping to do before I go on sabbatical this week; if we can for now run our internal refineries on k8s off this branch for this week before landing, then no rush at all.

TylerHelmuth: @RainofTerra can we do that?

TylerHelmuth: @lizthegrey if that is possible you should still update your branch to explicitly set DebugServiceAddr even if long term  the api for this feature in the helm chart changes.

lizthegrey: We'd need to publish a --devel version of the chart to the https://honeycombio.github.io/helm-charts/ repo. I can try to investigate.

lizthegrey: Pushed as v1.9.2-hnyinternal.0

lizthegrey: > @lizthegrey if that is possible you should still update your branch to explicitly set DebugServiceAddr even if long term the api for this feature in the helm chart changes.

Done in https://github.com/honeycombio/infra/pull/4432

TylerHelmuth: Added the debug section.

TylerHelmuth (approved):

akvanhar (approved):

charts/refinery/ci/debug.yaml
```diff
@@ -0,0 +1,11 @@
...
+  limits:
+    cpu: 100m
+    memory: 200M
+  requests:
+    cpu: 100m
+    memory: 200M
+
+debug:
+  enabled: true
+  port: 6062
```
akvanhar, line 11: Should this be 6061 to match values.yaml?

TylerHelmuth: No, this test is specifically showing that the value can be non-default and the chart will render.  I do need to update the default value to be 6060 tho.

PR #257, opened 2023-06-29
maint: Bump default refinery version to 2.0.0

TylerHelmuth:

- Updates the Refinery helm chart to use Refinery 2.0
- Closes https://github.com/honeycombio/helm-charts/issues/259

- Overhaul the values.yaml config and rule section to meet the new Refinery 2.0 configuration.  This is a breaking change.
- Update NOTES.txt to try and stop people from accidentally upgrading to this version of the chart.
- Remove outdated CI examples
- Update the README to be inline with 2.0 expectations.
- Remove sample config dir - with the new auto-generating docs we should link to those instead of keep old copies.

- CI tests
- locally in a kind cluster


cartermp (commented):
I don't know how helm actually works, but the bits in NOTES.txt seems like they're right and prevent people who haphazardly upgrade from getting a broken refinery

akvanhar (approved):
caveat: I know very little about helm charts. But things look good from my naive eyes.

mterhar (approved):
Tested to ensure it would error out with a refinery 1.x values file

PR #269, opened 2023-07-11
fix: Fix issue where default rule was always present unless explicitly disabled.

TylerHelmuth:

- The include of a default in the values.yaml led to some confusion.  In order to override the default rule you had to explicitly set `.Values.rules.Samplers.__default__.DeterministicSampler: null`, which, while 100% correct in helm, is a little unintuative.

- Updates the chart to only set a default rule if none is supplied by the user.  This ensures that Refinery always has a default rule, but won't mess with any users supplied configuration.

Tested locally and added a new CI scenario

### DIFF
--- a/charts/refinery/ci/rules-values.yaml
+++ b/charts/refinery/ci/rules-values.yaml
@@ -1,0 +1,17 @@
+resources:
+  limits:
+    cpu: 100m
+    memory: 200M
+  requests:
+    cpu: 100m
+    memory: 200M
+
+rules:
+  RulesVersion: 2
+  Samplers:
+    __default__:
+      DeterministicSampler:
+        SampleRate: 100
+    test:
+      DeterministicSampler:
+        SampleRate: 1
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -54,13 +54,14 @@ config:
 # See https://docs.honeycomb.io/manage-data-volume/refinery/sampling-methods/ for how to configure Refinery rules.
 rules:
   RulesVersion: 2
-  Samplers:
+  Samplers: {}
     # The default sampler is used when no other sampler matches.
     # It is required to have a default sampler.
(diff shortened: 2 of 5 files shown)


charts/refinery/values.yaml
```diff
@@ -54,13 +54,14 @@ config:
...
   RulesVersion: 2
-  Samplers:
+  Samplers: {}
     # The default sampler is used when no other sampler matches.
     # It is required to have a default sampler.
-    # The name consists of two underscores on either side of the word "default".
-    __default__:
-        DeterministicSampler:
-            SampleRate: 1
+    # If you do not supplie a default sampler the helm chart will inject
```
akvanhar, line 60:
```suggestion
    # If you do not supply a default sampler the helm chart will inject
```

charts/refinery/values.yaml
```diff
@@ -54,13 +54,14 @@ config:
...
+    # the DeterministicSampler show below.
```
kentquirk, line 61:
```suggestion
    # If you do not supply a default sampler the helm chart will inject
    # the DeterministicSampler shown below.
```

charts/refinery/ci/rules-values.yaml
```diff
@@ -0,0 +1,17 @@
...
+  requests:
+    cpu: 100m
+    memory: 200M
+
+rules:
+  RulesVersion: 2
+  Samplers:
+    __default__:
+      DeterministicSampler:
+        SampleRate: 100
```
kentquirk, line 14: Is 100 the value we want here?

kentquirk (approved):
Approved, but please accept my change.

charts/refinery/ci/rules-values.yaml
TylerHelmuth: This is for a test, proving that we can set a default sampler and it will merge correctly.  I chose a non-default value to show that the value really is changing.

PR #291, opened 2023-09-15
refinery/redis deployment: add ability to set inidividual pod annotations

NaurisSadovskis:
Ability to set pod annotations which is useful when configuring scraping configuration, such as DataDog ([example](https://docs.datadoghq.com/integrations/redisdb/?tab=kubernetes)).

This PR maintains backwards compatibility with existing annotations and merges with any optional values added as part of `values.redis.podAnnotations`.

### DIFF
--- a/charts/refinery/templates/deployment-redis.yaml
+++ b/charts/refinery/templates/deployment-redis.yaml
@@ -16,7 +16,8 @@ spec:
   {{- include "refinery.redis.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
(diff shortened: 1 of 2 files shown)


charts/refinery/templates/deployment-redis.yaml
```diff
@@ -16,7 +16,8 @@ spec:
   {{- include "refinery.redis.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
+      {{- $annotationsMerged := merge .Values.redis.podAnnotations .Values.podAnnotations }}
+      {{- with $annotationsMerged }}
```
TylerHelmuth, line 20: Sorry reviewing this took so long.  Can you move this logic into a helper template?

PR #292, opened 2023-09-26
feat: Add hny-network-agent chart

MikeGoldsmith:
Adds a helm chart for the new Honeycomb Network Agent.

- Closes https://github.com/honeycombio/honeycomb-network-agent/issues/87

- Adds new chart, values, and templated resources

The new agent can be installed with this chart.


MikeGoldsmith:
I believe the test is failing because the chart fails to install if neither honeycomb.apikey or an honeycomb.existingSecret is set.

charts/hny-network-agent/templates/NOTES.txt
```diff
@@ -0,0 +1,18 @@
...
+  / // /  / |/ / \ \/ /       / |/ / ___  / /_ _    __ ___   ____  / /__       / _ |  ___ _ ___   ___  / /_
+ / _  /  /    /   \  /       /    / / -_)/ __/| |/|/ // _ \ / __/ /  '_/      / __ | / _ `// -_) / _ \/ __/
+/_//_/  /_/|_/    /_/       /_/|_/  \__/ \__/ |__,__/ \___//_/   /_/\_\      /_/ |_| \_, / \__/ /_//_/\__/ 
+                                                                                    /___/                  
+
+
+Honeycomb Network Agent is setup and configured to collect events from your Kubernetes cluster.
+You should see data flowing within a few minutes at https://ui.honeycomb.io.
+
+{{ if .Values.honeycomb.apiKey }}
```
MikeGoldsmith, line 12: @TylerHelmuth I think we've lost our validation rules for checking the following conditions:
- Neither honeycomb.apiKey or honeycomb.existingSecret is set
- Both honeycomb.apiKey or honeycomb.existingSecret are set
- honeycomb.existingSecret is set and honeycomb.existingSecretKey is not set

TylerHelmuth:
@MikeGoldsmith we need to add a `ci` folder with a `default-values.yaml` file that sets `honeycomb.apiKey` to some bogus value. Here is what we do for the honeycomb kubernetes agent:

```
honeycomb:
  apiKey: example-api-key
```

charts/network-agent/templates/daemonset.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "hny-network-agent.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
```
TylerHelmuth, line 6: We need to add these labels to all `metadata.labels` (only the labels at the root of the objects): https://github.com/open-telemetry/opentelemetry-helm-charts/blob/8402aeb044cf47f8f47590a3f094fdb79395a745/charts/opentelemetry-collector/templates/_helpers.tpl#L52-L62

charts/network-agent/templates/secret.yaml
TylerHelmuth, line 1: I would prefer we do not make a secret for users, instead depending on them to create a secret or inserting the API key directly.  I believe the work for this, and the warning, are back in the other PR.

MikeGoldsmith: I didn’t create the secret at first like in the other PR but found the user experience felt weird.

I think creating the secret to hold the API key if passed directly and include messaging to indicate the implications has a better user experience. Some security is better than none, right?

The readme, notes.txt and values.yaml all include details on what to expect. 

If we don’t want to manage the secret, I’d prefer to remove the option to pass in the API key directly and just expect a secret to be passed in all the time.

TylerHelmuth: I definitely see the value in the ergonomics of passing in the apiKey and making a secret for the user.  Another drawback, though, is that you either have to put the apiKey in the values.yaml file in plain text or in the install command via a `--set`. Although the method promotes security on the installed resources, it does not on the actual install steps.

With `--set` you can at least use and env var to hide the value, which is how we should endorse this functionality.

MikeGoldsmith: Yep, I agree setting using —set should use an env var. We do that in the agent readme. I’ll update the examples here to do the same. 

Maybe we suggest anyone using values.yaml to also use  secret to avoid putting their API key in plain text?

Are you okay keeping the secret if we do the example and recommendation above?

TylerHelmuth: Yes

charts/network-agent/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+name: hny-network-agent
+description: Honeycomb Network Agent
+version: 0.0.1
+appVersion: v0.0.17-alpha
```
JamieDanielson, line 5:
```suggestion
appVersion: v0.0.19-alpha
```

charts/network-agent/README.md
```diff
@@ -0,0 +1,45 @@
+# Honeycomb Network Agent
+
+The Helm chart install the [Honeycomb Network Agent](https://github.com/honeycombio/honeycomb-network-agent) in a Kubernetes cluster.
+
+## TL;DR;
+
+```bash
```
JamieDanielson, line 7:
```suggestion
```console
```

I would usually use shell for all of these, but it looks like console is the one consistently used the rest of this file and in the honeycomb readme at least.

charts/network-agent/README.md
```diff
@@ -0,0 +1,45 @@
+# Honeycomb Network Agent
+
+The Helm chart install the [Honeycomb Network Agent](https://github.com/honeycombio/honeycomb-network-agent) in a Kubernetes cluster.
+
+## TL;DR;
```
JamieDanielson, line 5:
```suggestion
## TL;DR
```

JamieDanielson (commented):
This looks good! A few suggestions on the first pass-through with the easy happy path of setting api key directly, confirmed the unpublished version here can be installed with the following:

```sh
export HONEYCOMB_API_KEY=<mykey>
helm install hny-network-agent --set honeycomb.apiKey=$HONEYCOMB_API_KEY ./charts/network-agent
```

charts/network-agent/templates/daemonset.yaml
MikeGoldsmith: Added helpers to manage common and selector labels in https://github.com/honeycombio/helm-charts/pull/292/commits/c322ed778a57a87443e2a0d2c3ffbb962602a895

MikeGoldsmith:
I think this ready for another (final?) review, thanks for your feedback 👍🏻 

cc @TylerHelmuth @JamieDanielson

charts/network-agent/values.yaml
```diff
@@ -0,0 +1,19 @@
+honeycomb:
```
JamieDanielson, line 1: Do we want to add the debug and log level options or are we intentionally omitting them for the first pass?

MikeGoldsmith: I was going to say omit for now, but they definitely add value and will likely be more useful during initial feedback. I'll add them 👍🏻

charts/network-agent/templates/daemonset.yaml
```diff
@@ -0,0 +1,43 @@
...
+            - name: HONEYCOMB_API_ENDPOINT
+              value: {{ .Values.honeycomb.endpoint }}
+            - name: HONEYCOMB_DATASET
+              value: {{ .Values.honeycomb.dataset }}
+            - name: HONEYCOMB_STATS_DATASET
+              value: {{ .Values.honeycomb.statsDataset }}
+          securityContext:
+            capabilities:
+              add:
+                - NET_RAW
```
TylerHelmuth, line 43: Not a blocker for this PR, but It would be a good idea to add a section in the README detailing the permissions/RBAC this chart sets up.

TylerHelmuth (approved):
Left a couple nits, but nothing worth blocking anymore.

charts/network-agent/templates/daemonset.yaml
MikeGoldsmith: We don't have a section in the main repo that describes required permissions yet. We've got the following ticket to track document them, and I've added a note to say we should update the helm chart once we've got it
- https://github.com/honeycombio/honeycomb-network-agent/issues/167

charts/network-agent/values.yaml
MikeGoldsmith: I've added a simpler solution of allow extra env vars to be set and passed into the daemonset in https://github.com/honeycombio/helm-charts/pull/292/commits/c0545bde0c1debfec2c2d02e75a687453af14e17. Our and the OTel charts do the same thing.

I like this better because it doesn't let some accidentally to mess around with debug / development settings. They need to know what to set.

PR #296, opened 2023-10-05
maint: [network-agent] fix up network-agent chart

JamieDanielson:

[Failed release](https://app.circleci.com/pipelines/github/honeycombio/helm-charts/819/workflows/01cf638e-3eef-4593-81f9-757850d900dc/jobs/431) because Chart name (hny-network-agent) didn't match directory name (network-agent). 

- We opted for network-agent for chart name because honeycomb and/or honeycombio will be included when using the chart. The dataset names and container/pod names are intentionally kept as hny-network-agent to make it more easily distinguishable since these are things the user will see in a running k8s cluster. 
- Added the environment variables necessary for the use of k8s downward API in the agent, which adds k8s metadata to events so the agent knows things about itself.

Events have all expected k8s metadata... and the release works 😅 😅

### DIFF
--- a/charts/network-agent/Chart.yaml
+++ b/charts/network-agent/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: hny-network-agent
+name: network-agent
 description: Honeycomb Network Agent
 version: 0.0.1
 appVersion: v0.0.20-alpha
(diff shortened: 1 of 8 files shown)


robbkidd (approved):

robbkidd:
But hold up, we're going to try this locally.

robbkidd:
Tested locally. Hangs together. 🦖 🌮 👍🏻

TylerHelmuth (approved):

PR #297, opened 2023-10-06
feat: [network-agent] add resource limits

JamieDanielson:

- Closes https://github.com/honeycombio/honeycomb-network-agent/issues/273

- Add `resources` template to daemonset to allow end users to set resource limits
- Move `extraEnvVars` to bottom of daemonset
- Add `resources` to `values.yaml`

Installing the helm chart with resource limits defined should set those limits as expected

### DIFF
--- a/charts/network-agent/templates/daemonset.yaml
+++ b/charts/network-agent/templates/daemonset.yaml
@@ -37,9 +37,6 @@ spec:
               value: {{ .Values.honeycomb.dataset }}
             - name: HONEYCOMB_STATS_DATASET
               value: {{ .Values.honeycomb.statsDataset }}
-            {{- with .Values.extraEnvVars }}
-            {{- toYaml . | nindent 12 }}
-            {{- end }}
             # https://kubernetes.io/docs/concepts/workloads/pods/downward-api/
             - name: AGENT_NODE_IP
               valueFrom:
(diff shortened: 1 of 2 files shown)


charts/network-agent/templates/daemonset.yaml
```diff
@@ -61,6 +58,11 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+            {{- with .Values.extraEnvVars }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          resources:
```
TylerHelmuth, line 64: We should only render `resources:` if the user has set `.Values.resources`.  You can do something like `{{- if .Values.resources }}`

JamieDanielson: Oh good catch! Updated in https://github.com/honeycombio/helm-charts/pull/297/commits/9dacca5c255ab74f97f4039c5861846ecab7c52c and after this update I can see in a `--dry-run` that it no longer shows in the final daemonset.

TylerHelmuth (approved):

TylerHelmuth:
It is also a good idea to add a warning in NOTES.txt when resources is not set. The community collector helm chart has an example.

MikeGoldsmith (approved):

TylerHelmuth (approved):

PR #300, opened 2023-10-15
maint: Bump agent version to v0.0.22-alpha

MikeGoldsmith:
Updates agent image version to [v0.0.22-alpha](https://github.com/honeycombio/honeycomb-network-agent/releases/tag/v0.0.22-alpha)

- Update chart's appVersion to v0.0.22-alpha
- Updates additional env var examples

### DIFF
--- a/charts/network-agent/Chart.yaml
+++ b/charts/network-agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: network-agent
 description: Honeycomb Network Agent
 version: 0.0.1
-appVersion: v0.0.20-alpha
+appVersion: v0.0.22-alpha
 home: https://honeycomb.io
 sources:
   - https://github.com/honeycombio/honeycomb-network-agent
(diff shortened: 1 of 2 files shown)


JamieDanielson (commented):
Looks like we added one of the new variables already in another PR but this would be a good time to add the others now available (feel free to grab this [commit 40c6f88](https://github.com/honeycombio/helm-charts/commit/40c6f88028c5f102de981738d8c39eeff9d7fb3f))

JamieDanielson:
> Looks like we added one of the new variables already in another PR but this would be a good time to add the others now available (feel free to grab this [commit 40c6f88](https://github.com/honeycombio/helm-charts/commit/40c6f88028c5f102de981738d8c39eeff9d7fb3f))

Now that I'm looking at this, I think we should update the headers example here too, especially because it's case-sensitive and User-Agent is something we would want to encourage and it's dropped without it being specified here. What do you think of something like this that I'm adding to the agent repo now: `value: "User-Agent,Header-1,header-2"`

JamieDanielson (approved):

TylerHelmuth (approved):

PR #301, opened 2023-10-15
rel: Prepare v0.0.2 of network agent chart

MikeGoldsmith:
Prepares the v0.0.22-alpha release of the network agent. 

- Update chart version to v0.0.2
- Add changelog entry

Note:
This PR is based on the following PR and needs to wait for it to merge first:
- https://github.com/honeycombio/helm-charts/pull/300

### DIFF
--- a/charts/network-agent/CHANGELOG.md
+++ b/charts/network-agent/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Honeycomb Network Agent Helm Chart Changelog
 
+## Honeycomb Network Agent v0.0.2
+
+- feat: Add resource limits (#297) | @JamieDanielson
+- maint: Adds HTTP headers example to values.yaml (#299) | @MikeGoldsmith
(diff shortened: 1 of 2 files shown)


JamieDanielson (commented):
I merged the other PR, this has the extra commit with the appVersion update so not sure if that should be removed first or if it doesn't matter because it gets squashed anyway.
Does show a merge conflict though, can you take a look at why that would be a problem?

MikeGoldsmith:
> I merged the other PR, this has the extra commit with the appVersion update so not sure if that should be removed first or if it doesn't matter because it gets squashed anyway. Does show a merge conflict though, can you take a look at why that would be a problem?

No idea why, but it saw the chart's version bump as a conflict 🤷🏻 Ready for another review.

JamieDanielson (approved):

TylerHelmuth (approved):

PR #307, opened 2023-11-02
maint: [network-agent] reduce apiGroups in rbac

JamieDanielson:

unnecessary extra lookups

- remove apps and metrics from rbac

everything should still work as well as it did before


puckpuck (approved):
Perfect!

TylerHelmuth (approved):

PR #308, opened 2023-11-17
feat: add deploymentLabels to refinery chart

niels-s:
<!--
Thank you for contributing to the project! 💜
Please see our [OSS process document](https://github.com/honeycombio/home/blob/main/honeycomb-oss-lifecycle-and-practices.md#) to get an idea of how we operate.
-->

In the same fashion as podLabels. Add the ability to configure custom deployment labels for the refinery helm chart.


When running the following commands, the output should contain a `Deployment` resource for Refinery which contains the customer label `example.com/label: "true"`

```
echo 'deploymentLabels:                                                                                                                                                                                                                                                                                                                                                                                            
  example.com/label: "true"' > values.yml
helm template refinery charts/refinery --namespace=test --values=values.yml
```

### DIFF
--- a/charts/refinery/templates/deployment.yaml
+++ b/charts/refinery/templates/deployment.yaml
@@ -5,6 +5,9 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
   {{- include "refinery.labels" . | nindent 4 }}
+  {{- with .Values.deploymentLabels }}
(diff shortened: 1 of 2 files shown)


TylerHelmuth (approved):
Thanks for this addition.  As a heads up, we probably wont release this chart until the next refinery release: https://github.com/honeycombio/refinery/milestone/6

PR #315, opened 2023-11-27
fix: update honeycomb templates to use release namespace

### LINKED ISSUE #158: Add namespace support to Honeycomb helm chart

fchikwekwe:

This updates the honeycomb helm chart to use release namespace just like the collector helm chart did. 

Tested changes locally to ensure that namespace changes still work as intended. 
```
➜  kubectl config set-context --current --namespace=default    
Context "kind-kind" modified.
➜  helm template ./charts/opentelemetry-collector | grep -i namespace
WARNING: This chart is deprecated
  namespace: default
  namespace: default
  namespace: default
  namespace: default
  namespace: default
➜ kubectl config set-context --current --namespace=kube-public
Context "kind-kind" modified.
➜  helm template ./charts/opentelemetry-collector | grep -i namespace
WARNING: This chart is deprecated
  namespace: kube-public
  namespace: kube-public
  namespace: kube-public
  namespace: kube-public
  namespace: kube-public
```

- Closes #158

### DIFF
--- a/charts/honeycomb/templates/cluster-role.yaml
+++ b/charts/honeycomb/templates/cluster-role.yaml
@@ -3,6 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: {{ include "honeycomb.agent.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "honeycomb.agent.labels" . | nindent 4 }}
   annotations:
--- a/charts/honeycomb/templates/rbac.yaml
+++ b/charts/honeycomb/templates/rbac.yaml
@@ -3,6 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: {{ include "honeycomb.agent.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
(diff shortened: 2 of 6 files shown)


charts/honeycomb/templates/cluster-role.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: {{ include "honeycomb.agent.fullname" . }}
+  namespace: {{ .Release.Namespace }}
```
TylerHelmuth, line 6:
```suggestion
```
ClusterRole is a non-namespaced resource so we can omit this addition.

charts/honeycomb/templates/rbac.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: {{ include "honeycomb.agent.fullname" . }}
+  namespace: {{ .Release.Namespace }}
```
TylerHelmuth, line 6:
```suggestion
```
ClusterRoleBinding is also non-namespaced

TylerHelmuth (approved):

TylerHelmuth (approved):

PR #319, opened 2023-12-01
maint: bump network-agent to v0.1.0-alpha

robbkidd:

Updating network-agent chart to install latest version, v0.1.0-beta.

- bump the version of the network-agent installed by this chart

### DIFF
--- a/charts/network-agent/Chart.yaml
+++ b/charts/network-agent/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: network-agent
 description: Honeycomb Network Agent
-version: 0.0.8
+version: 0.1.0
 appVersion: v0.1.0-beta
 home: https://honeycomb.io
 sources:


akvanhar (approved):

charts/network-agent/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: network-agent
 description: Honeycomb Network Agent
-version: 0.0.7
-appVersion: v0.0.25-alpha
+version: 0.0.8
```
TylerHelmuth, line 4: Do you want to do a minor version bump in the chart, or are you waiting until the app version drop's beta?

robbkidd: 🤔 I hadn't thought about it.

The _chart_ didn't change much, so I bumped patch. I suppose we could take the chart to 0.1.0 as well, but I don't have a well-informed opinion about chart versioning.

TylerHelmuth: Normally if the app version takes a minor bump the chart should too since it's providing new features via the app

TylerHelmuth: Sounds like this would be a good time to do our first 0.1.0 release of the chart.

robbkidd: Okeydoke!

TylerHelmuth (approved):

PR #324, opened 2023-12-06
feat(network-agent): Add pod annotation for secret checksum

### LINKED ISSUE #323: Add config checksum pod annotation for generated secret to network agent chart
It is common practice to add a checksum value of the values file used to configure the helm chart as a pod label. This allows values file changes trigger an update of pods to use the new values.

MikeGoldsmith:
When the agent is configured with an API key, it automatically creates a secret for it to be stored in and is used to pass in the key to the agent pod via env var. However, if the secret is changed on a running deployment, k8s doesn't consider the pod state to have changed because the env var is injected from the secret. 

This PR adds the checksum of the generated secret file as a pod annotation, which will trigger a redeployment of pods if the API key in the secret changes.

- Closes #323

- Add checksum of generated secret file if the API key is passed in via values file
- Add note to readme when using an existing secret to add a pod annotation to help trigger redeployments if the value changes.

A checksum pod annotation is added to daemonset pods when using an API key via values file.

Example helm output when using dry run:
```
...
# Source: network-agent/templates/daemonset.yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: network-agent
  namespace: default
  labels:
    app.kubernetes.io/name: network-agent
    app.kubernetes.io/instance: network-agent
    app.kubernetes.io/version: "v0.1.0-beta"
    app.kubernetes.io/managed-by: Helm
    helm.sh/chart: network-agent-0.1.0
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: network-agent
      app.kubernetes.io/instance: network-agent
  template:
    metadata:
      annotations:
        checksum/secret: 09162b2af9ca1d481a18d2d350eb4d94c331155ae3bd19f6ecc178a494e0a44c
      labels:
        app.kubernetes.io/name: network-agent
        app.kubernetes.io/instance: network-agent
    spec:
...
```

### DIFF
--- a/charts/network-agent/README.md
+++ b/charts/network-agent/README.md
@@ -39,7 +39,7 @@ helm install network-agent honeycomb/network-agent --set honeycomb.existingSecre
 
 The default key used to retrieve the API from the secret key is `apiKey`. You can provide an alternative key by setting the `existingSecretKey` value.
 
-**NOTE**: If you use a separate secret key to manage your API keys, it is also recommened to set a custom pod annotation that can be used to trigger a redeployment of pods.
+**NOTE**: If you use a separate secret key to manage your API keys, you should also define a custom pod annotation (for example, a hash of the secret) that can be used to trigger a redeployment of pods when the secret changes.
 
 ### Using a values.yaml
 


charts/network-agent/README.md
```diff
@@ -39,6 +39,8 @@ helm install network-agent honeycomb/network-agent --set honeycomb.existingSecre
 
 The default key used to retrieve the API from the secret key is `apiKey`. You can provide an alternative key by setting the `existingSecretKey` value.
 
+**NOTE**: If you use a separate secret key to manage your API keys, it is also recommened to set a custom pod annotation that can be used to trigger a redeployment of pods.
```
kentquirk, line 42:
```suggestion
**NOTE**: If you use a separate secret key to manage your API keys, you should also define a custom pod annotation (for example, a hash of the secret) that can be used to trigger a redeployment of pods when the secret changes.
```

charts/network-agent/README.md
```diff
@@ -39,6 +39,8 @@ helm install network-agent honeycomb/network-agent --set honeycomb.existingSecre
...
+
 ### Using a values.yaml
 
 If you prefer to manage your helm charts by providing your own values.yaml, it is recommended to use an existing secret to avoid storing your Honeycomb API key as plain text.
```
kentquirk, line 46: I know this isn't part of this changeset, but why do we use the term "existing secret" here? As opposed to a new secret? I don't understand what this is supposed to mean, but I could be missing something.

kentquirk (approved):
🦖

PR #332, opened 2024-01-11
feat: Define redis resources

### LINKED ISSUE #331: Refinery - Redis container resources
It is usually recommended for each container in a pod to have proper resources requests and limits. While these can be set for refinery itself, I could not find a way to set resources for the redis deployment used for peer discovery.

I believe redis resources should be configurable through values.yaml

gjacquet:

It should be possible to set redis resources from values.yaml

- Closes #331

Allow setting redis resources from values.

Configure resources in values.yaml under `redis.resources` and see that redis container resources are set accordingly.

### DIFF
--- a/charts/refinery/Chart.yaml
+++ b/charts/refinery/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 2.5.0
+version: 2.5.1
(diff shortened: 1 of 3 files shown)


TylerHelmuth (commented):
@gjacquet looks good.  Please bump the patch version of the chart.

gjacquet:
Done!

TylerHelmuth (approved):

PR #336, opened 2024-01-11
feat: Create PDB for refinery.

### LINKED ISSUE #333: [refinery] Create PDB for refinery
Currently when there is a cluster / node maintenance, there is no way to control how many refinery pods will remain available.

It should be possible to define a PodDisruptionBudget to control disruption in the Helm chart

gjacquet:

Allows creating a Pod Disruption Budget for refinery
- Closes #333

Add template and configuration to create a PDB for refinery (disabled by default)

Set `podDisruptionBudget.enabled: true` and check PDB has been created.


MikeGoldsmith (approved):
I've played around with this and it looks good - thanks @gjacquet 👍🏻

charts/refinery/values.yaml
```diff
@@ -255,6 +255,13 @@ tolerations: []
 
 affinity: {}
 
+podDisruptionBudget:
```
cartermp, line 258: Should this have a comment documenting it?

charts/refinery/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 2.5.1
+version: 2.5.2
```
TylerHelmuth, line 5: This is a "big" feature add, lets bump the minor version.

charts/refinery/values.yaml
TylerHelmuth: Agreed.  Please add a comment above `podDisruptionBudget` describing what this section enables and when someone might what to enable it.

charts/refinery/values.yaml
```diff
@@ -255,6 +255,13 @@ tolerations: []
...
+  enabled: false
+  labels: {}
+  annotations: {}
+  minAvailable:
```
TylerHelmuth, line 262: Lets comment this out with a note that the field is allowed, but not at the same time as `maxUnavailable`.  Then lets add a check to `NOTES.txt` that both are not set at the same time.

charts/refinery/templates/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,26 @@
...
+  {{- with .Values.podDisruptionBudget.minAvailable }}
+  minAvailable: {{ . }}
+  {{- end }}
+  {{- with .Values.podDisruptionBudget.maxUnavailable }}
+  maxUnavailable: {{ . }}
+  {{- end }}
+  selector:
+    matchLabels:
+      {{- include "refinery.selectorLabels" . | nindent 6 }}
+{{- end }}
```
TylerHelmuth, line 26:
```suggestion
{{- end }}

```

charts/refinery/values.yaml
gjacquet: Done

charts/refinery/values.yaml
gjacquet: Added comment

TylerHelmuth: @gjacquet please also comment out `minAvailable` in the vaules.yaml and add a new section in `NOTES`.txt that will fail the install if both `minAvailable` and `maxUnavailable` are set.  I believe this is necessary because `PodDisruptionBudget` doesn't allow both being set at the same time.

gjacquet: Done

charts/refinery/templates/NOTES.txt
```diff
@@ -1,2 +1,19 @@
...
+
+PDB can be enabled and configured through chart values. E.g.
+
+```yaml
+podDisruptionBudget:
+  enabled: true
+  # minAvailable and maxUnavailable cannot be both specified at the same time.
+  # minAvailable:
+  maxUnavailable: 1
+```
```
TylerHelmuth, line 16:
```suggestion
```

charts/refinery/values.yaml
```diff
@@ -271,13 +271,13 @@ podDisruptionBudget:
...
   # evictions by specifying "100%".
-  minAvailable:
+  # minAvailable:
 
   # An eviction is allowed if at most "maxUnavailable" pods selected by
   # "selector" are unavailable after the eviction, i.e. even in absence of the
   # evicted pod. For example, one can prevent all voluntary evictions by
   # specifying 0. This is a mutually exclusive setting with "minAvailable".
-  maxUnavailable: 1
+  # maxUnavailable: 1
```
TylerHelmuth, line 280:
```suggestion
  maxUnavailable: 1
```

charts/refinery/templates/NOTES.txt
```diff
@@ -1,2 +1,19 @@
...
+{{- if and .Values.podDisruptionBudget.minAvailable .Values.podDisruptionBudget.maxUnavailable }}
+  {{- fail "PDB minAvailable and maxUnavailable cannot be both specified at the same time." }}
```
TylerHelmuth, line 18:
```suggestion
  {{- fail "podDisruptionBudget.minAvailable and podDisruptionBudget.maxUnavailable cannot be both specified at the same time." }}
```

TylerHelmuth (approved):

TylerHelmuth (commented):
@gjacquet I'm sorry but I have 1 more request.  Can you please add a new values.yaml file in the `charts/refinery/ci` folder that enables this new feature?

TylerHelmuth (approved):

PR #337, opened 2024-01-12
feat: Create refinery headless service.

### LINKED ISSUE #335: [refinery] Headless service for refinery
We have noticed that all our otel collectors tend to send spans to only a single refinery instance. This typically occurs after a refinery rolling restart where otel collector tends to establish connection to a single pod, usually one of the first one that was created during the restart.

This describes the issue and the solution pretty well: https://kubernetes.io/blog/2018/11/07/grpc-load-balancing-on-kubernetes-without-tears/

When using grpc to send spans to refinery, it is usually better to use a headless service for refinery and do client side load balancing in otel collector.

gjacquet:
We have noticed that all our otel collectors tend to send spans to only a single refinery instance. This typically occurs after a refinery rolling restart where otel collector tends to establish connection to a single pod, usually one of the first one that was created during the restart.

- Closes #335

Allow creating a headless service for refinery.

Set `headlessService.enabled: true` and check headless service was created.


TylerHelmuth (commented):
Instead of a new service template, could we achieve the same result if we allowed the user to specify `ClusterIP` in the existing `service` section?

gjacquet:
@TylerHelmuth that would be more a question for you with your knowledge of all the use cases for refinery.

I have seen different charts maintainer go different routes. E.g Bitnami's Thanos chart has a whole set of services to accommodate all possible use cases within a single deployment: https://github.com/bitnami/charts/tree/main/bitnami/thanos/templates/query, on the other hand, otel collector has only a single service that can optinally be headless: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/service.yaml#L19.

The driving factor would be how many different client you expect and how they expect load balancing to be handled.

If there will for sure be only one client type, e.g. all client side load balanced grpc, then single service is good. If you expect a more heterogeneous setup with a mix client side load balanced clients for gprc and kube proxy balanced http clients then it would be better to support both cluster IP and headless.

TylerHelmuth:
For now, since the Refinery deployment options aren't very complex, I think a single service with some configuration options is the best play.  Let's implement something similar to the OpenTelemetry Collector chart where `ClusterIp` is a configurable option on the `service` section.

gjacquet:
removed the extra service and added support for the cluster IP under the existing service config

TylerHelmuth (approved):
thank you!

PR #338, opened 2024-01-15
feat: Configure refinery labels and annotations for each resources

### LINKED ISSUE #334: [refinery] Resource labels
It would be nice to be able to define custom labels on all resources managed by this chart.

Values file should support extra entries for each resources to define extra labels

gjacquet:

It would be nice to be able to define custom labels on all resources managed by this chart.

- Closes #334

Allows defining a set of common labels to be applied to all resources.
It also enables managing labels and annotations on each resources independently.

E.g. set `extraLabels: <some labels>`  or `configMap.labels: <labels>` and verify epxected labels have been added to corresponding resources.

### DIFF
--- a/charts/refinery/templates/deployment-redis.yaml
+++ b/charts/refinery/templates/deployment-redis.yaml
@@ -17,8 +17,8 @@ spec:
   template:
     metadata:
       {{- /*
-        FIXME .Values.podAnnotations should not be used at all here.
-        Keeping it for backward compatibiltiy but should eventually be removed.
+       .Values.podAnnotations is used here to maintain backwards compatibility.
+       It should be removed eventually.
       */}}
       {{- if or .Values.podAnnotations .Values.redis.podAnnotations }}
       annotations:
--- a/charts/refinery/templates/deployment-redis.yaml
+++ b/charts/refinery/templates/deployment-redis.yaml
@@ -6,9 +6,6 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
   {{- include "refinery.redis.labels" . | nindent 4 }}
-  {{- with .Values.redis.annotations }}
-    annotations: {{ toYaml . | nindent 4 }}
-  {{- end }}
 {{- with .Values.redis.annotations }}
   annotations: {{ toYaml . | nindent 4 }}
 {{- end }}
@@ -19,6 +16,10 @@ spec:
   {{- include "refinery.redis.selectorLabels" . | nindent 6 }}
   template:
     metadata:
+      {{- /*
+        FIXME .Values.podAnnotations should not be used at all here.
+        Keeping it for backward compatibiltiy but should eventually be removed.
+      */}}
       {{- if or .Values.podAnnotations .Values.redis.podAnnotations }}
       annotations:
       {{- with .Values.podAnnotations }}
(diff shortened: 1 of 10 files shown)


charts/refinery/templates/deployment-redis.yaml
```diff
@@ -16,12 +19,23 @@ spec:
   {{- include "refinery.redis.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
+      {{- if or .Values.podAnnotations .Values.redis.podAnnotations }}
       annotations:
+      {{- with .Values.podAnnotations }}
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
```
gjacquet, line 26: I am not entirely sure why we would want to use `.Values.podAnnotations` for redis pods but keeping this for backward compatibility.

TylerHelmuth: Can you add a comment above the condition specifying this backwards compatibility as the reason why both are used?

charts/refinery/templates/deployment-redis.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
   {{- include "refinery.redis.labels" . | nindent 4 }}
+  {{- with .Values.redis.annotations }}
+    annotations: {{ toYaml . | nindent 4 }}
+  {{- end }}
```
TylerHelmuth, line 11: These lines are duplicates of the 3 lines below them.

charts/refinery/templates/service-redis.yaml
```diff
@@ -4,8 +4,15 @@ kind: Service
 metadata:
   name: {{ include "refinery.redis.fullname" . }}
   namespace: {{ .Release.Namespace }}
+  {{- with .Values.redis.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   labels:
     {{- include "refinery.redis.labels" . | nindent 4 }}
+    {{- with .Values.redis.labels }}
```
TylerHelmuth, line 13: I think this needs to be `.Values.redis.extraLabels`.

TylerHelmuth (commented):
Due to the similarities I am having a tough time reviewing this PR.  I think we'd benefit from breaking this up into smaller chunks, probably by k8s object being modified.

gjacquet:
@TylerHelmuth sure, we can take baby steps.
I created #339 to support common labels, this will cleanup this diff a bit once merged.

charts/refinery/templates/deployment-redis.yaml
gjacquet: Done

charts/refinery/templates/deployment-redis.yaml
```diff
@@ -16,15 +16,27 @@ spec:
   {{- include "refinery.redis.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
+      {{- /*
+        FIXME .Values.podAnnotations should not be used at all here.
+        Keeping it for backward compatibiltiy but should eventually be removed.
+      */}}
```
TylerHelmuth, line 22:
```suggestion
      {{- /*
       .Values.podAnnotations is used here to maintain backwards compatibility.
       It should be removed eventually.
      */}}
```

I _think_ that comment won't get rendered.

TylerHelmuth (approved):

PR #339, opened 2024-01-17
feat: Add refinery common labels.

gjacquet:

It would be nice to be able to define custom labels on all resources managed by this chart.

- Contributes to #334

Define values to be used as a set of common labels to be applied on all
refinery and redis resources

Set `commonLabels: <some labels>` or `redis.commonLabels: <some labels>`  and verify epxected labels have been added to corresponding resources.

### DIFF
--- a/charts/refinery/templates/deployment.yaml
+++ b/charts/refinery/templates/deployment.yaml
@@ -37,6 +37,9 @@ spec:
       {{- with .Values.podLabels }}
       {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.redis.commonLabels}}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
     spec:
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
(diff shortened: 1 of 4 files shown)


charts/refinery/templates/deployment.yaml
```diff
@@ -37,6 +37,9 @@ spec:
       {{- with .Values.podLabels }}
       {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.redis.commonLabels}}
```
TylerHelmuth, line 40: I don't think we should be adding the redis labels to the Refinery pod

gjacquet: my bad, that was a bad copy paste

TylerHelmuth (commented):
The git compare is weird, but I believe we need to bump to `2.6.2`

gjacquet:
yup, some conflict needed resolving. Should be good now

TylerHelmuth (approved):

PR #342, opened 2024-01-18
feat: Refinery lifecycle

### LINKED ISSUE #341: [refinery] Add lifecycle configuration to container
To make rolling update smoother, it is very common to add a sleep to containers preStop lifecycle to give Kubernetes some time to reconfigure the application's service endpoints.

You can find more details there: https://blog.gruntwork.io/delaying-shutdown-to-wait-for-pod-deletion-propagation-445f779a8304

It should be possible to configure each container's lifecycle to accomodate the use case decribed above but this could possibly be used in other cases too.

**Describe alternatives you've considered**

**Additional context**

gjacquet:

To make rolling update smoother, it is very common to add a sleep to containers preStop lifecycle to give Kubernetes some time to reconfigure the application's service endpoints.

- Closes #341

Allow configuring each container lifecycle.

Set values:
``` yaml
lifecycle: {}
    preStop:
      exec:
        command:
          - "sh"
          - "-c"
          - "sleep 10"
```
and check lifecycle has been configured as expected

### DIFF
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -133,6 +133,16 @@ environment: []
   #       name: honeycomb
   #       key: api-key
 
+# Configure container lifecycle hooks.
+# This can be used to add a prestop hook to make rolling restarts more robust for instance.
+lifecycle: {}
+  # preStop:
+  #   exec:
+  #     command:
+  #       - "sh"
+  #       - "-c"
+  #       - "sleep 10"
+
 # Use to map additional volumes into the Refinery pods
 # Useful for volume-based secrets
 extraVolumeMounts: []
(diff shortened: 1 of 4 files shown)


charts/refinery/values.yaml
```diff
@@ -133,6 +133,16 @@ environment: []
...
 
+# Configure container lifecycle hooks.
+# This can be used to add a prestop hook to make rolling restarts more robust for instance.
+lifecycle: {}
+  # preStop:
+  #   exec:
+  #     command:
+  #       - "sh"
+  #       - "-c"
+  #       - "sleep 10"
```
TylerHelmuth, line 144:
```suggestion
lifecycle: {}
```

I'd prefer not to suggest anything here since we use the `scratch` image and can't really run commands.

charts/refinery/values.yaml
```diff
@@ -177,6 +187,16 @@ redis:
...
+  # Configure container lifecycle hooks.
+  # This can be used to add a prestop hook to make rolling restarts more robust for instance.
+  lifecycle: {}
+    # preStop:
+    #   exec:
+    #     command:
+    #       - "sh"
+    #       - "-c"
+    #       - "sleep 10"
+
```
TylerHelmuth, line 199:
```suggestion
  lifecycle: {}
  
```

TylerHelmuth (approved):

PR #355, opened 2024-03-13
maint: Update ubuntu image in workflows to latest

MikeGoldsmith:
Older ubuntu images used in our workflows are being marked as deprecated. We need to update to newer ones.

- Update workflows images to use `ubuntu-2204:2024.01.1` image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,7 @@
 version: 2
 jobs:
   test-modified-charts:
-    machine:
-      image: ubuntu-2004:202201-02
+    runs-on: ubuntu-latest
     environment:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -25,8 +25,7 @@ jobs:
 
 
   release-charts:
-    machine:
-      image: ubuntu-2004:202201-02
+    runs-on: ubuntu-latest
```
TylerHelmuth, line 28: Can we use the same version for both? Either latest or specific is fine with me

MikeGoldsmith: Oops - accidentally left one when trying something. Both use the same image now 👍🏻

PR #360, opened 2024-05-09
[refinery] Make ShutdownDelay based on terminationGracePeriodSeconds

TylerHelmuth:

Refinery 3 attempts to preserve data on shutdown, and has a configurable ShutdownDelay field where it tries to process remaining work while shutting down. In k8s we need this config option to be inline with `terminationGracePeriodSeconds` to ensure k8s doesn't kill pods early.

- added terminationGracePeriodSeconds configuration option in values.yaml
- set `ShutdownDelay` to be `terminationGracePeriodSeconds - 5` seconds by default.  Users can override this value in their values.yaml

Tested locally in kind.

### DIFF
--- a/charts/refinery/Chart.yaml
+++ b/charts/refinery/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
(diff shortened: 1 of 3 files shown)


kentquirk (approved):
Make it so

PR #363, opened 2024-05-16
feat(refinery): Allow configuring probes

TylerHelmuth:

Allows users to configure readiness and liveness probes values

- allows setting `initialDelaySeconds`, `periodSeconds`, `timeoutSeconds`, and `failureThreshold` for `livenessProbe` and `readinessProbe`.

Tested locally


kentquirk (approved):
LGTM

PR #377, opened 2024-07-23
feat(refinery): enable templating for liveness and readiness probe

VinozzZ:

Refinery 2.7 has set up health check endpoints for readiness and liveness probes from k8s

- templating liveness and readiness probe in Refinery Deployment spec

local testing

### DIFF
--- a/charts/refinery/values.yaml
+++ b/charts/refinery/values.yaml
@@ -104,6 +104,22 @@ resources:
     cpu: 500m
     memory: 500Mi
 
+# liveness probe configuration
+# Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+livenessProbe:
+  initialDelaySeconds: 10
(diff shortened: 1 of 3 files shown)


charts/refinery/values.yaml
```diff
@@ -104,6 +104,22 @@ resources:
...
+  timeoutSeconds: 1
+  failureThreshold: 3
+
+# readiness probe configuration
+# Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+readinessProbe:
+  initialDelaySeconds: 0
+  periodSeconds: 1
+  timeoutSeconds: 1
+  failureThreshold: 3
```
kentquirk, line 121: I think we want failure threshold for ready to be 1 -- if it ever responds not ready, it's not ready. I might add like a 3 second initialDelay, though, just so we don't time out the first request.

kentquirk (approved):

PR #382, opened 2024-09-03
feat(refinery): Set ShutdownDelay based on terminationGracePeriodSeconds

VinozzZ:

Refinery 2.8 attempts to drain data to remaining peers on shutdown, and has a configurable ShutdownDelay field where it tries to process remaining work while shutting down. In k8s we need this config option to be inline with terminationGracePeriodSeconds to ensure k8s doesn't kill pods early.

it's the same change as #360 
- added terminationGracePeriodSeconds configuration option in values.yaml
- set ShutdownDelay to be terminationGracePeriodSeconds - 5 seconds by default. Users can override this value in their values.yaml
- publish a `-hnyinternal` tag for the change

Test localling using kind

### DIFF
--- a/charts/refinery/Chart.yaml
+++ b/charts/refinery/Chart.yaml
@@ -3,7 +3,7 @@ name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
 version: 2.11.2-hnyinternal.0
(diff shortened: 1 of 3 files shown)


kentquirk (approved):
LGTM!

PR #390, opened 2024-09-19
feat(refinery): add support for topologySpreadConstraints

### LINKED ISSUE #97: Consider Enabling `topologySpreadConstraints` for Refinery Helm Chart
I've seen behavior from the Refinery chart where pods will be scheduled on the same Kubernetes Node

We've seen good success using `topologySpreadConstraints` to spread workload across K8s nodes using a `topologyKey` like `topology.kubernetes.io/zone`. Allowing operators to choose a `topologyKey` would allow us to mandate a more fault-tolerant HA topology.

nakamume:

- Closes #97 

Adds support for providing `topologySpreadConstraints` field. This will help us deploy refinery confirming to some topology constraints.

### DIFF
--- a/charts/refinery/Chart.yaml
+++ b/charts/refinery/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 2.11.4
+version: 2.12.0
 appVersion: 2.8.2
 keywords:
   - refinery
--- a/charts/refinery/Chart.yaml
+++ b/charts/refinery/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 2.11.3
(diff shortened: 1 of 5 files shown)


nakamume:
@TylerHelmuth can we get some ❤️ ❤️ here. Thanks.

JamieDanielson:
Thanks for this @nakamume ! At a glance this seems reasonable but I'm looking into whether there's any reason we haven't done it yet (is it just time/resources or something else?). I'd think this should be a feature (feat) not a chore, with a minor bump instead of a patch. And it looks like this would close #97 so we can link that up as well.

charts/refinery/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 2.11.3
+version: 2.11.4
```
JamieDanielson, line 5:
```suggestion
version: 2.12.0
```

PR #399, opened 2024-12-03
maint(refinery): update appVersion to 2.9

VinozzZ:

Update Refinery to the latest 2.9 release

### DIFF
--- a/charts/refinery/Chart.yaml
+++ b/charts/refinery/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 2.12.3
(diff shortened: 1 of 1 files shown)


charts/refinery/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 2.12.3
-appVersion: 2.8.4
+version: 2.12.4
```
TylerHelmuth, line 5: Let's bump the minor version

VinozzZ: Is it because this Refinery release is a minor release with bigger changes?

TylerHelmuth (approved):

PR #400, opened 2024-12-05
feat(refinery): add deployment-marker support

VinozzZ:

It's helpful to know when a new version of Refinery has been deployed when viewing Refinery's telemetry data. This PR adds a deployment marker job in Refinery's chart. The job will create a honeycomb marker when a Refinery deployment happenes.

NOTE: This is a MVP version of this feature. There could be other improvements made down the line

- add a deployment-marker job. The job utilize helm hooks to only run after a `helm install` or `helm upgrade`
- introduce a new `REFINERY_HONEYCOMB_DEPLOY_MARKER_API_KEY` environment variable for sending deploy markers to honeycomb

Tested it locally

### DIFF
--- a/charts/refinery/templates/deployment-marker-job.yaml
+++ b/charts/refinery/templates/deployment-marker-job.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.deployMarker.enabled }}
 apiVersion: batch/v1
 kind: Job
 metadata:
@@ -14,18 +15,27 @@ spec:
     spec:
       serviceAccountName: {{ include "refinery.serviceAccountName" . }}
       restartPolicy: Never
+      {{- with .Values.deployMarker.volumes }}
+      volumes:
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
       containers:
         - name: {{ .Chart.Name }}-deploy-marker
           image: "{{ .Values.deployMarker.image.repository }}:{{ .Values.deployMarker.image.tag }}"
           imagePullPolicy: {{ .Values.deployMarker.image.pullPolicy }}
 
-          {{- with .Values.environment }}
+          {{- with .Values.deployMarker.environment }}
           env:
+          {{- toYaml . | nindent 12 }}
+          {{- end }}
+
+          {{- with .Values.deployMarker.volumeMounts }}
+          volumeMounts:
           {{- toYaml . | nindent 12 }}
           {{- end }}
           args:
             - -k
-            - "$(REFINERY_HONEYCOMB_API_KEY)"
+            - "$(REFINERY_HONEYCOMB_DEPLOY_MARKER_API_KEY)"
             - -d
             - {{ .Values.deployMarker.honeycombDataset }}
             - --api_host
@@ -35,3 +45,4 @@ spec:
             - "Refinery {{ .Values.image.tag | default .Chart.AppVersion }}, Chart {{ .Chart.Version }}"
             - -t
             - deploy
+{{- end }}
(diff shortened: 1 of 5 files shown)


charts/refinery/templates/deployment-marker-job.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: batch/v1
```
TylerHelmuth, line 1: We should wrap all these contents in a check for `{{- if .Values.deployMarker.enabled }}`

charts/refinery/values.yaml
```diff
@@ -384,3 +384,15 @@ debug:
...
+
+deployMarker:
+  enabled: false
+  image:
+    repository: honeycombio/honeymarker
+    pullPolicy: IfNotPresent
+    # Overrides the image tag whose default is the chart appVersion.
+    tag: "latest"
+  honeycombAPI: ""
+  honeycombDataset: ""
```
TylerHelmuth, line 397: The mnost future proof approach is to require via a `{{- fail }} in NOTES.txt that if `.Values.deployMarker.enabled` then `.Values.deployMarker.honeycombAPI` and `.Values.deployMarker.honeycombDataset` are set. Then in the future if we want to provide a default value from `.Values.config` or something we wont be breaking any existing users.

charts/refinery/templates/deployment-marker-job.yaml
```diff
@@ -0,0 +1,37 @@
...
+      restartPolicy: Never
+      containers:
+        - name: {{ .Chart.Name }}-deploy-marker
+          image: "{{ .Values.deployMarker.image.repository }}:{{ .Values.deployMarker.image.tag }}"
+          imagePullPolicy: {{ .Values.deployMarker.image.pullPolicy }}
+
+          {{- with .Values.environment }}
+          env:
+          {{- toYaml . | nindent 12 }}
+          {{- end }}
```
TylerHelmuth, line 25: We should add a `.Values.deployMarker.env` to use instead so this job doesn't get any extra env vars.

charts/refinery/values.yaml
```diff
@@ -391,8 +391,11 @@ deployMarker:
   image:
     repository: honeycombio/honeymarker
     pullPolicy: IfNotPresent
-    # Overrides the image tag whose default is the chart appVersion.
+    # Overrides the image tag whose default is latest
     tag: "latest"
   honeycombAPI: ""
   honeycombDataset: ""
+  environment: {}
```
TylerHelmuth, line 399:
```suggestion
  # For now we require an env var name `REFINERY_HONEYCOMB_DEPLOY_MARKER_API_KEY` to contain the api key
  environment: {}
```

Can you include a commented out example of setting the `REFINERY_HONEYCOMB_DEPLOY_MARKER_API_KEY` value with a secret.  I think we do this other places in this file

VinozzZ: Are you referring to the commented out code on line 400?

TylerHelmuth: I mean lets add something like this:

```environment: []
  # - name: REFINERY_HONEYCOMB_DEPLOY_MARKER_API_KEY
  #   valueFrom:
  #     secretKeyRef:
  #       name: honeycomb
  #       key: api-key
```

VinozzZ: Yes, that's what I have on line 401

charts/refinery/values.yaml
```diff
@@ -384,3 +384,24 @@ debug:
   # Only used if debug is enabled.
   # This value will be used to set config.DebugServiceAddr if it is not already set.
   port: 6060
+
+
```
TylerHelmuth, line 388:
```suggestion
# This feature is EXPERIMENTAL and subject to breaking changes
#
# When enabled, the chart will use a post-install job to create a honeycomb marker
# in the specified dataset to mark the install of the chart.
```

TylerHelmuth (approved):

PR #401, opened 2024-12-11
feat(refinery): add template (tpl) support to extraCommandArgs

TylerHelmuth:

Refinery supports layering of config/rules and has the capability to get both from a URL. When using this chart as a subchart it is not possible to pass in additional config/rules using a url that is based on the parent chart unless this chart supports templating the input. 

Add `tpl` support to `extraCommandArgs`

Tested locally.  Adding

```yaml
extraCommandArgs:
  - "{{ .Release.Name }}"
```

results in 

```
command:
  - refinery
  - -c
  - /etc/refinery/config.yaml
  - -r
  - /etc/refinery/rules.yaml
  - testing
```

Non-templated strings are unaffected.


TylerHelmuth:
Converting to draft until I figure out a way to do this in older versions of helm. tpl inside range isn’t supported til helm 3.14

TylerHelmuth:
The test failed bc of my invalid logger api key value, but the templating did work as I hoped:

```
    Command:
      refinery
      -c
      /etc/refinery/config.yaml
      -r
      /etc/refinery/rules.yaml
      --logger-api-key
      refinery-7ctmc6izzq
```

So tpl in `with` is ok.

charts/refinery/ci/debug-values.yaml
codeboten, line 1: why the rename?

TylerHelmuth: I realized this test wasn't being executed bc it didn't end in `values.yaml`.  With the rename now it actually gets run by our CI.

codeboten (approved):

PR #403, opened 2024-12-12
feat(observability-pipeline): add control plane support

TylerHelmuth:

- Add support for installing a control plane

- Add template files for control plane
- Add extra args to refinery command to reach the control plane
- Updated readme
- [BREAKING] switch collectors to be disabled by default
- [BREAKING] send refinery metrics directly to honeycomb instead of to collectors by default.

Tested locally in kind.


charts/observability-pipeline/README.md
```diff
@@ -13,7 +13,8 @@ Create a Kubernetes secret to store your API Key
 
 ```shell
 export HONEYCOMB_API_KEY=mykey
-kubectl create secret generic honeycomb --from-literal=api-key=$HONEYCOMB_API_KEY
+export HONEYCOMB_MGMT_API_SECRET=mymanagementkey
+kubectl create secret generic honeycomb-observability-pipeline --from-literal=api-key=$HONEYCOMB_API_KEY --from-literal=management-api-secret=$HONEYCOMB_MGMT_API_SECRET
```
MikeGoldsmith, line 17: Should we do the same for the API key too?

MikeGoldsmith (approved):

charts/observability-pipeline/README.md
TylerHelmuth: the honeycomb management api key id is not considered a secret so I made a section in the `.Values.controlPlane` to set it.

TylerHelmuth: The `HONEYCOMB_API_KEY` referenced here is used by the collectors and refinery

PR #406, opened 2024-12-13
feat(observability-pipeline): allow setting existing serviceAccount name

TylerHelmuth:

Allows users to disable creating a service account for the control plane and instead specify the name of an existing service account to use.

- add serviceAccount subsection to controlPlane
- add logic to use the serviceAccount.name if it is set 

Tested locally


codeboten (approved):
🎉

PR #408, opened 2024-12-18
fix(refinery): deployment marker image pull secrets and job name uniqueness

VinozzZ:

- Allow refinery pods to authenticate and pull honeymarker image.
- Add uniqueness for deployment marker job name so that concurrent deployment won't fail

- use the same `imagePullSecrets` from refinery for deployment marker
- add a randomly generated string to deployment marker job name

tested locally

### DIFF
--- a/charts/refinery/templates/deployment-marker-job.yaml
+++ b/charts/refinery/templates/deployment-marker-job.yaml
@@ -2,7 +2,7 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ include "refinery.fullname" . }}-deploy-marker
+  name: {{ include "refinery.fullname" . }}-deploy-marker-{{ randAlphaNum 5 | lower }}
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "refinery.labels" . | nindent 4 }}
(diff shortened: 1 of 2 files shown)


charts/refinery/templates/deployment-marker-job.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ include "refinery.fullname" . }}-deploy-marker
+  name: {{ include "refinery.fullname" . }}-deploy-marker-{{ randAlphaNum 5 | lower }}
```
MikeGoldsmith, line 5: What happens if the generated value was previously used? Does it care as long as it's not currently running?

Is there a k8s job number or something we could hook into like we get for CI piplines?

MikeGoldsmith (approved):
Looks good. Left a query regarding the random value but not blocking.

robbkidd (approved):
dancing marker

charts/refinery/templates/deployment-marker-job.yaml
VinozzZ: I tried to use the `generateName` but helm isn't happy with that. This job is suppose to be deleted as soon as it finishes. The randomness for this name doesn't need to be too good.

ferd (approved):
sometimes this stuff is funny, just make sure to look at the cluster 1h post deployment to know if the markers still work or if the marker pods are crashing/failing to come up.

PR #412, opened 2025-01-16
feat: add HTP helm chart

TylerHelmuth:

- Adds a new chart for deploying Honeycomb Telemetry Pipeline

- add new htp chart
- include readme instructions for install
- include NOTES.txt to enforce secrets
- This chart requires a LICENSE to install so we skip the `ct install` test in CI

Test locally using kind both install paths.


JamieDanielson (approved):
🚀

PR #421, opened 2025-01-24
feat(observability-pipeline): use pvc for cacheing latest config

VinozzZ:

- adds support for using pvc to persist latest config in control plane

- add a new `persistentVolumnClaimName` in the `values.yaml` file to allow users to mount a volume
- pass in the mount path using `-config-cache-path` flag

### DIFF
--- a/charts/observability-pipeline/templates/control-plane-deployment.yaml
+++ b/charts/observability-pipeline/templates/control-plane-deployment.yaml
@@ -25,10 +25,14 @@ spec:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.controlPlane.image.repository }}:{{ .Values.controlPlane.image.tag }}"
           imagePullPolicy: {{ .Values.controlPlane.image.pullPolicy }}
-          {{- if .Values.controlPlane.telemetry.enabled }}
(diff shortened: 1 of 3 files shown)


charts/observability-pipeline/templates/control-plane-deployment.yaml
```diff
@@ -25,11 +25,15 @@ spec:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.controlPlane.image.repository }}:{{ .Values.controlPlane.image.tag }}"
           imagePullPolicy: {{ .Values.controlPlane.image.pullPolicy }}
-          {{- if .Values.controlPlane.telemetry.enabled }}
           args:
+          {{- if .Values.controlPlane.telemetry.enabled }}
```
TylerHelmuth, line 29: if neither of these conditions are true then we'll render an empty `args` which I am pretty sure is an issue.

VinozzZ: I added a condition to wrap the whole `args` section

TylerHelmuth (approved):

PR #424, opened 2025-02-05
fix(htp): fix bug where username/password could not come from secret

TylerHelmuth:

- fixes an issue where the default username and password in the values.yaml were taking priority over any username/password set in the `config.secret`. This is an outcome of the bindplane chart: https://github.com/observIQ/bindplane-op-helm/blob/main/charts/bindplane/templates/bindplane.yaml#L135-L154

- remove default username and password values.
- update readme instructions

tested locally

### DIFF
--- a/charts/htp/Chart.yaml
+++ b/charts/htp/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: htp
 description: Honeycomb Telemetry Pipeline
 type: application
-version: 0.1.0
+version: 0.1.1
 appVersion: 1.85.0
 home: https://honeycomb.io
 sources:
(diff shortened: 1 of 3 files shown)


MikeGoldsmith (approved):

JamieDanielson (approved):
I think this needs docs changes as well

robbkidd:
> I think this needs docs changes as well

[Yope](https://docs.honeycomb.io/send-data/telemetry-pipeline/install/#install-with-kubernetes-using-helm)

verythorough:
Docs PR here: https://github.com/honeycombio/docs/pull/2502 (Thanks, @TylerHelmuth!)

PR #432, opened 2025-03-05
feat(observability-pipeline): set otel resource attributes

TylerHelmuth:

Since internal telemetry is not flowing through a k8sattributes processor use the downward api to set some useful k8s resource attributes instead

https://app.asana.com/0/1208469935858869/1209582431136181

- update collector container to set some k8s attributes
- update beekeeper container to set some k8s attributes

Tested locally

image


codeboten (approved):

charts/observability-pipeline/templates/beekeeper-deployment.yaml
```diff
@@ -47,10 +65,6 @@ spec:
               value: {{ .Values.beekeeper.pipelineInstallationID }}
             - name: HONEYCOMB_MGMT_API_KEY
               value: {{ .Values.beekeeper.publicMgmtKey }}
-            {{- if .Values.beekeeper.telemetry.enabled }}
```
VinozzZ, line 50: Is there any downside for removing this condition to only enable the env var if telemetry is enabled?

PR #435, opened 2025-03-06
feat(observability-pipeline): Update default env vars to be a map

TylerHelmuth:

Lists of env vars are hard to change because they can't be merged. Updated collector and beekeeper's default env vars to be a map so individual entries can easily be disabled or changed

- https://app.asana.com/0/1208469935858869/1209598578859275

- reorganize values.yaml
- update beekeeper and collector env var list in deployments
- removed un-needed check in NOTE.txt
- add new extraEnvs option for beekeeper and collector to allow addition env vars

Local templating

### DIFF
--- a/charts/observability-pipeline/values.yaml
+++ b/charts/observability-pipeline/values.yaml
@@ -34,21 +34,30 @@ beekeeper:
   team: ""
   publicMgmtKey: ""
   persistentVolumeClaimName: ""
-  environment:
-    - name: HONEYCOMB_MGMT_API_SECRET
-      valueFrom:
-        secretKeyRef:
-          name: honeycomb-observability-pipeline
-          key: management-api-secret
-    - name: HONEYCOMB_API_KEY
-      valueFrom:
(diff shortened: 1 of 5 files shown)


charts/observability-pipeline/values.yaml
```diff
@@ -34,21 +34,30 @@ beekeeper:
...
-        secretKeyRef:
-          name: honeycomb-observability-pipeline
-          key: api-key
-    - name: TELEMETRY_ENDPOINT
-      value: https://api.honeycomb.io:443
-    - name: TELEMETRY_KEY
-      value: "${env:HONEYCOMB_API_KEY}"
+  defaultEnv:
+    HONEYCOMB_MGMT_API_SECRET:
+      enabled: true
```
VinozzZ, line 39: why do we have a `enabled` option for this if beekeeper requires this environment variable to be set in order to start?

VinozzZ (approved):

PR #437, opened 2025-03-06
feat(observability-pipeline): allow configuring metrics dataset

TylerHelmuth:

https://app.asana.com/0/1208469935858869/1209598568289018

- adds support for new `TELEMETRY_METRICS_DATASET` env var to allow configuring how beekeeper injects the metrics dataset header for its clients.

- add new env var
- add ability to set collector's `service.name` value

Tested locally in kind

image

### DIFF
--- a/charts/observability-pipeline/templates/_helpers.tpl
+++ b/charts/observability-pipeline/templates/_helpers.tpl
@@ -109,7 +109,7 @@ agent:
   executable: /otelcol-contrib
   description:
     identifying_attributes:
-      service.name: collector
+      service.name: {{ .Values.collector.serviceName }}
 
 
 storage:
--- a/charts/observability-pipeline/templates/_helpers.tpl
+++ b/charts/observability-pipeline/templates/_helpers.tpl
@@ -109,7 +109,7 @@ agent:
   executable: /otelcol-contrib
   description:
     identifying_attributes:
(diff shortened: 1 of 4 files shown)


TylerHelmuth:
Draft until new beekeeper image is available

charts/observability-pipeline/templates/_helpers.tpl
```diff
@@ -109,7 +109,7 @@ agent:
   executable: /otelcol-contrib
   description:
     identifying_attributes:
-      service.name: collector
+      service.name: {{ .Values.collector.serviceName }}
```
VinozzZ, line 112: I don't think beekeeper currently supports custom service.name for its agents. https://github.com/honeycombio/beekeeper/blob/05cac45b6e09f529f065e72e793cfa28e8e640e4/internal/opamp/agent_type.go#L22

TylerHelmuth: Ah, so if the chart makes this configurable we could break interaction with Beekeeper?

VinozzZ: Yea, I believe beekeeper rely on this to differentiate which config sent to which agent

codeboten (approved):

PR #460, opened 2025-04-10
feat(refinery): Make it easy to set eu1 urls

TylerHelmuth:

Right now eu customer need to manually set a couple refinery configuration endpoints to point to eu honeycomb instead of prod. That is kind of annoying. This PR allows the chart to automatically set those fields based on a flag.

- add new `region` setting that will automatically configure refinery's endpoints based on the value
- Add comments

tested locally using `helm template`

### DIFF
--- a/charts/refinery/templates/_helpers.tpl
+++ b/charts/refinery/templates/_helpers.tpl
@@ -204,15 +204,15 @@ OTelTracing:
 
 {{- define "refinery.localConfig" -}}
 Network:
-  HoneycombAPI: http://localhost:8081
+  HoneycombAPI: http://host.docker.internal:8081
 HoneycombLogger:
-  APIHost: http://localhost:8081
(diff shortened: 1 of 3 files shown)


charts/refinery/templates/_helpers.tpl
```diff
@@ -119,6 +137,84 @@ Build config file for Refinery
...
+LegacyMetrics:
+  APIHost: https://api-kibble.honeycomb.io
+OTelMetrics:
+  APIHost: https://api-kibble.honeycomb.io
+OTelTracing:
+  APIHost: https://api-kibble.honeycomb.io
+{{- end }}
+
+{{- define "refinery.localConfig" -}}
+Network:
```
JamieDanielson, line 206: Does anything else need to change to allow non-secure traffic?

JamieDanielson (approved):

PR #464, opened 2025-04-16
feat: update observability pipeline chart refinery options

codeboten:

This updates the refinery configuration options for the experimental opamp support.

### DIFF
--- a/charts/observability-pipeline/Chart.yaml
+++ b/charts/observability-pipeline/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: observability-pipeline
 description: Chart to deploy both OpenTelemetry Collector and Honeycomb Refinery
 type: application
-version: 0.0.39-alpha
+version: 0.0.38-alpha
 appVersion: 0.0.1-alpha
 keywords:
   - refinery
--- a/charts/observability-pipeline/Chart.yaml
+++ b/charts/observability-pipeline/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: observability-pipeline
 description: Chart to deploy both OpenTelemetry Collector and Honeycomb Refinery
 type: application
-version: 0.0.38-alpha
+version: 0.0.39-alpha
 appVersion: 0.0.1-alpha
 keywords:
   - refinery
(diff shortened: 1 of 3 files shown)


charts/observability-pipeline/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: observability-pipeline
 description: Chart to deploy both OpenTelemetry Collector and Honeycomb Refinery
 type: application
-version: 0.0.37-alpha
+version: 0.0.39-alpha
```
codeboten, line 5:
```suggestion
version: 0.0.38-alpha
```

charts/observability-pipeline/values.yaml
```diff
@@ -7,12 +7,13 @@ global:
...
   image:
     repository: "honeycombio/refinery"
-    tag: "2.9.3-htp-dev.1"
+    tag: "2.9.4-htp-dev.1"
   config:
     Network:
-      OpAMPEnabled: true
-      OpAMPEndpoint: "ws://{{ include \"honeycomb-observability-pipeline.name\" . }}-beekeeper:4320/v1/opamp"
-      OpAMPRecordUsage: false
+    OpAMP:
```
TylerHelmuth, line 13: the network section can be removed

charts/observability-pipeline/values.yaml
```diff
@@ -7,12 +7,13 @@ global:
 refinery:
   image:
     repository: "honeycombio/refinery"
-    tag: "2.9.3-htp-dev.1"
+    tag: "2.9.4-htp-dev.1"
   config:
     Network:
```
codeboten, line 12:
```suggestion
```

charts/observability-pipeline/values.yaml
codeboten: Removed

TylerHelmuth (approved):

TylerHelmuth (changes requested):
You need to update the schema file too with the new opamp setting. That'll take a change in the refinery chart.

charts/refinery/values.schema.json
TylerHelmuth, line 1: Lets do this in a separate PR, will also need a chart version patch value bump

TylerHelmuth (approved):

PR #468, opened 2025-04-21
feat(observability-pipeline): add json schema for the chart

TylerHelmuth:

Related to https://github.com/honeycombio/pipeline-team/issues/239

Add a json schema to help enforce keys in the values.yaml. This ensures users don't mistype something and get confused. It will also help with changing values in later versions if needed.

Tested locally with `helm template`.

### DIFF
--- a/charts/observability-pipeline/values.yaml
+++ b/charts/observability-pipeline/values.yaml
@@ -1,9 +1,46 @@
+# The ID of your pipeline installation.
+# This can be found on the Pipeline Installations page in the Honeycomb UI.
 pipelineInstallationID: ""
+
+# The public ID of your Honeycomb Management API key.
+# See https://docs.honeycomb.io/configure/teams/manage-api-keys/#find-management-api-keys for more details.
 publicMgmtAPIKey: ""
 
 global:
+  # This setting configures the default endpoints. All endpoints can still be configured and will take precedence.
+  # The default value is '' (unset). When unset, U.S. values will be used for any unconfigured endpoints.
+  # Other valid values are:
+  #  - production-eu
+  #
+  # When 'production-eu' is set, the following settings are configured:
+  #  refinery:
+  #    config:
+  #      Network:
+  #        HoneycombAPI: https://api.eu1.honeycomb.io
+  #      HoneycombLogger:
+  #        APIHost: https://api.eu1.honeycomb.io
+  #      LegacyMetrics:
+  #        APIHost: https://api.eu1.honeycomb.io
+  #      OTelMetrics:
+  #        APIHost: https://api.eu1.honeycomb.io
+  #      OTelTracing:
+  #        APIHost: https://api.eu1.honeycomb.io
+  #  beekeeker:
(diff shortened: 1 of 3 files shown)


charts/observability-pipeline/values.yaml
```diff
@@ -143,13 +214,20 @@ primaryCollector:
...
             key: api-key
   extraEnvs: []
+
   volumes: []
   volumeMounts: []
+
+  # configuration options for the OpAMP Supervisor.
   opampsupervisor:
     telemetry:
+      # Enablesthe OpAMP Supervisor's internal telemetry.
```
JamieDanielson, line 224:
```suggestion
      # Enables the OpAMP Supervisor's internal telemetry.
```

charts/observability-pipeline/values.yaml
```diff
@@ -26,17 +63,18 @@ refinery:
...
-  resources: {}
-  serviceAccount:
-    create: true
-    name: ""
   image:
     repository: "honeycombio/beekeeper"
     pullPolicy: IfNotPresent
     tag: "latest"
+
+  # The Hnoeycomb API endpoint Beekeeper should use.
```
JamieDanielson, line 73:
```suggestion
  # The Honeycomb API endpoint Beekeeper should use.
```

charts/observability-pipeline/values.yaml
```diff
@@ -1,9 +1,46 @@
...
+  #        HoneycombAPI: https://api.eu1.honeycomb.io
+  #      HoneycombLogger:
+  #        APIHost: https://api.eu1.honeycomb.io
+  #      LegacyMetrics:
+  #        APIHost: https://api.eu1.honeycomb.io
+  #      OTelMetrics:
+  #        APIHost: https://api.eu1.honeycomb.io
+  #      OTelTracing:
+  #        APIHost: https://api.eu1.honeycomb.io
+  #  beekeeker:
```
JamieDanielson, line 28:
```suggestion
  #  beekeeper:
```

JamieDanielson (commented):
I love the comments 🤩  I just had a few typo suggestions on the first pass. I was trying to test things out but we have a bug with the default region that is making it difficult (separate ticket created).

JamieDanielson:
Also I'm probably showing my lack of experience building helm charts but can you clarify how the json schema was created? Is it something that needs to get updated regularly? If we use a tool, maybe we can note that somewhere and throw it into the releasing or developing docs. What do you think?

TylerHelmuth:
> Is it something that needs to get updated regularly? If we use a tool, maybe we can note that somewhere and throw it into the releasing or developing docs. What do you think?

Manually. It mostly needs to stay up to date when we add new root objects.

JamieDanielson (approved):
I've brushed up my knowledge of these schema files and this looks good 😄 

To confirm it did what I expected, I changed a property type in the `values.schema.json` file and ran `helm lint .` within the observability-pipeline directory. This resulted in an error that I expected to see (`Invalid type. Expected: string, given: object`). 🚀

PR #490, opened 2025-05-29
feat: Added template rendering tests

martinjt:

When committing changes that will change the rendered output, it's hard to know if there are adverse effects to that.

This PR adds 2 testing scripts.

`render-all-test-templates.sh` which will render all the charts based on values files located in the tests folder. This also strips the details that always change like the chart version to make it clearer.
`compare-test-rendered.sh` which will compare 2 sets of rendered files.

I've also added this to CI, so that if someone has not committed the updated rendered files CI will fail. This will ensure that PR reviews show all the expected changes a customer will see.

The build should pass, this PR makes no changes to the charts.

### DIFF
--- a/tests/render-all-test-templates.sh
+++ b/tests/render-all-test-templates.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+set -e
+
+BUILD_DEPENDENCIES=false
+RENDERED_DIR=rendered
+
+while getopts "bd:" opt; do
+  case $opt in
+    b) BUILD_DEPENDENCIES=true;;
+    d) RENDERED_DIR=$OPTARG;;
+    \?)
+      echo "Invalid option: -$OPTARG" >&2
+      exit 1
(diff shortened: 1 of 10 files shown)


tests/render-all-test-templates.sh
```diff
@@ -0,0 +1,54 @@
...
+done
+
+chart_dirs=($(ls -d */))
+
+for chart_dir in "${chart_dirs[@]}"; do
+
+    if [ $BUILD_DEPENDENCIES = true ]; then
+        echo "Building dependencies"
+        for chart_dir in */; do
+            helm repo add honeycomb https://honeycombio.github.io/helm-charts
```
TylerHelmuth, line 26: Does this fail if the repo is already added?

TylerHelmuth: confirmed it doesnt

TylerHelmuth (approved):

PR #501, opened 2025-06-04
maint: move htp to htp-bindplane

TylerHelmuth:

- Closes https://github.com/honeycombio/pipeline-team/issues/420

Rename the `htp` chart to `htp-bindplane`. The original `htp` chart will be deleted some time in the future.

- duplicate `htp` into `htp-bindplane`.
- update `htp` readme with sunset message

local install

### DIFF
--- a/charts/htp-builder/Chart.lock
+++ b/charts/htp-builder/Chart.lock
@@ -1,6 +1,0 @@
-dependencies:
-- name: refinery
-  repository: https://honeycombio.github.io/helm-charts
-  version: 2.17.0
-digest: sha256:13978ebf7edcd1c49266c9c1597442ce2a4a85774702048efd16ecd5738c333d
-generated: "2025-06-04T11:54:30.458257-06:00"
--- a/charts/htp-builder/Chart.yaml
+++ b/charts/htp-builder/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: htp-builder
 description: Chart to deploy Honeycomb Telemetry Pipeline Builder
 type: application
-version: 0.0.60-alpha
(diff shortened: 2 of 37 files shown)


charts/htp/README.md
```diff
@@ -1,60 +1,5 @@
...
-```sh
-kubectl port-forward svc/htp 3001
-```
-
-By default the helm chart sets the username to `admin` and the password to `admin`.
-
-## Values
-
-See the [subchart's values.yaml](https://github.com/observIQ/bindplane-op-helm/blob/main/charts/bindplane/values.yaml) for details.
+If you are a user of this chart, please switch to the [`htp-bindplane` chart](../htp-bindplane) instead. The `values.yaml` structure is the same, so no need to change your `values.yaml`, only the helm chart you're installing.
```
codeboten, line 5:
```suggestion
Existing users of this chart, please switch to the [`htp-bindplane` chart](../htp-bindplane) instead. The `values.yaml` structure is the same, so no need to change your `values.yaml`, only the helm chart being installed.
```

codeboten (approved):

Makefile
```diff
@@ -4,7 +4,7 @@ lint:
 
 .PHONY: test
 test:
-	ct install --config ./ct.yaml --excluded-charts htp --excluded-charts htp-builder
+	ct install --config ./ct.yaml --excluded-charts htp --excluded-charts htp-builder --excluded-charts htp-builder
```
codeboten, line 7:
```suggestion
	ct install --config ./ct.yaml --excluded-charts htp --excluded-charts htp-builder --excluded-charts htp-bindplane
```

PR #502, opened 2025-06-04
maint: rename observability-pipeline to htp-builder

TylerHelmuth:

- Closes https://github.com/honeycombio/pipeline-team/issues/421

- rename chart to htp-builder
- update readme
- update default secret name
- update template names
- update notes.txt (calvin s ascii)

sippycup


charts/htp-builder/Chart.yaml
```diff
@@ -12,6 +12,8 @@ keywords:
   - sampling
   - tracing
   - collector
+  - htp
+  - builder
```
codeboten, line 16: i wonder if builder is confusing in this context (with the existing ocb)

TylerHelmuth:
```suggestion
```

codeboten (approved):

PR #503, opened 2025-06-08
fix: move management api key into section

TylerHelmuth:

- Closes https://github.com/honeycombio/pipeline-team/issues/417

- moves mangement api key management into its own section
- add comments
- update NOTES.txt checks
- update json schema
- remove unneeded default env var
- update beekeeper template to use new section

test rendering


martinjt:
We should make sure there's an associated ticket in the Straws UI project to get the installation docs updated before Launch.

PR #518, opened 2025-06-26
Support image digests in refinery chart

jace-ys:

This aims to extend the refinery chart to support image digests, allowing one to specify a specific digest (e.g. `sha256:06a3f7b12a56ed051612416d3083814e145afc7fc3fc52a67aeffa348a20383c`) to ensure immutability of the image being deployed.

- Adds additional templates for images used in the refinery chart to `_helpers.tpl`
- Replaces the image templating in each deployment to use the templates in `_helpers.tpl`

- Template out the chart with digests set (e.g. `sha256:06a3f7b12a56ed051612416d3083814e145afc7fc3fc52a67aeffa348a20383c`) in values.yaml


jace-ys:
Thanks @TylerHelmuth!

I didn't update any docs or changelogs though, so if you could help with that that'll be appreciated 🙈

PR #534, opened 2025-09-30
maint(refinery): remove BufferSizes from config schema

VinozzZ:

In the upcoming Refinery 3 release, Refinery has removed the need to configure buffer sizes.

remove `BufferSizes` from schema file

### DIFF
--- a/charts/refinery/values.schema.json
+++ b/charts/refinery/values.schema.json
@@ -75,9 +75,6 @@
         "Collection": {
           "type": "object"
         },
-        "BufferSizes": {
-          "type": "object"
-        },
         "Specialized": {
           "type": "object"
         },


charts/refinery/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: refinery
 description: Chart to deploy Honeycomb Refinery
 type: application
-version: 2.18.2
-appVersion: 2.9.7
+version: 2.19.0
```
robbkidd, line 5: 🤔 In the past, we have bumped the major version of the chart when the app major version changes. My first inclination is to continue doing that even though we know this chart will work with a 3.0.0 app image. If user's values YAML for config uses deprecated-and-removed options, their clusters will fail to start.

charts/htp-builder/values.yaml
```diff
@@ -67,7 +67,7 @@ refinery:
     repository: honeycombio/refinery
     pullPolicy: IfNotPresent
     # Overrides the image tag whose default is the chart appVersion.
-    tag: "2.9.7"
+    tag: "3.0.0"
```
robbkidd, line 70: I think this change to htp-builder means this chart needs a version bump, too.

I don't know if we do this in the same PR or do separate PRs. 🤔

VinozzZ: We have decided to not touch htp-builder

robbkidd:
Should we remove `LegacyMetrics` from the values.schema.json as well?

VinozzZ:
> Should we remove `LegacyMetrics` from the values.schema.json as well?

That's a good point! I just removed it

PR #537, opened 2025-10-07
fix(refinery): use correct link for migration doc in README

VinozzZ:

Use the correct doc link for migration doc in README.md


robbkidd (approved):
Yea, that's better.

PR #559, opened 2025-11-20
feat(htp-builder): support ingress for primaryCollector [PIPE-208]

TylerHelmuth:

Adds Ingress support to `htp-builder` primaryCollector following pattern from the open-telemetry/opentelemetry-collector helm chart.

Closes PIPE-208

- Add new ingress template to render an Ingress object
- update values.yaml to allow configuring an ingress for the primary collector
- update json schema

helm template locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,7 @@
 .vscode/
 *.code-workspace
 
+### Claude Code ###
+.claude/
+
 tests/*/rendered-compare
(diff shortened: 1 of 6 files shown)


verythorough:
Not sure why Linear didn't auto-link. Trying this way?

Closes https://linear.app/honeycombio/issue/PIPE-208/helm-chart-add-ingress-support-to-primary-collector-in-htp-builder

